PR #7, opened 2020-11-19
Migrate gtk use to GTK4

BrainBlasted:
Use GTK4 APIs for the window identifier.


bilelmoussaoui:
you can get a window handle under wayland by using something like this https://github.com/gtk-rs/gtk4-rs/pull/46#issuecomment-723469276

bilelmoussaoui:
I went ahead and redid this by keeping the gtk3 & gtk4 features separately. Thanks!

PR #9, opened 2021-02-14
update zbus

bilelmoussaoui:


bilelmoussaoui:
@sophie-h this might interest you as I managed to fix the zbus issue and progressed a bit with the port to zbus 2.0 beta. So you can finally use the Async variant of each proxy. What I'm wondering is whether it makes sense to provide some kind of simple functions wrappers around the various proxies (at least some of them) to simplify the usage. 

For example the pick_color function in ScreenshotProxy could use a new helper function that would look like this (pseudo code)
```rust
async fn pick_color<'a>(parms...) ->  Pin<Box<dyn Future<Output = zbus::Result<Response<Color>>> + 'a>> {
    let (sender, receiver) = futures_channel::oneshot::channel();
    let sender = Cell::new(Some(sender));

    let connection = zbus::azync::Connection::new_session().await?;
    let proxy = AsyncScreenshotProxy::new(&connection)?;
    let request_handle =
        proxy.pick_color(params...).await?;
    let request = AsyncRequestProxy::new_for_path(&connection, request_handle.as_str())?;
    request.connect_response(|response: Response<Color>| {
        let sender = sender.replace(None);
        let _ = sender.send(response);
        Box::pin(Ok(()))
    }).await?;

    Box::pin(async move {
        if let Ok(response) = receiver.await {
            Ok(response)
        } else {
            Ok(Response::Err(ResponseError::Other))
        }
    })
}
```
the interesting part is that we can easily generate those with a macro for a lot of the portals we have. 

Anyway, if you have a bit of time to test/play with it a bit, I would appreciate any feedback. Especially API-wise. Thanks!

sophie-h:
Actually the oneshot::channel part is something I would expect from a zbus async API. But I have little experience with dbus responses. Maybe the general case is more complex.

sophie-h:
Seems like zbus::azync::Proxy::call(...).await is what should work instead of the channel workaround.

sophie-h:
The working code version. I still think there should be a version of `connect_response` that just returns the response. The other thing that feels like boiler plate is `AsyncRequestProxy` but not sure how to avoid that, because `request_background()` is generated.

```rust
    let connection = ashpd::zbus::azync::Connection::new_session().await?;
    let proxy = background::AsyncBackgroundProxy::new(&connection)?;

    let request_handle = proxy
        .request_background(
            ashpd::WindowIdentifier::default(),
            background::BackgroundOptions::default().reason(&gettext(
                "Schedule and continue running backups in background.",
            )),
        )
        .await?;

    let request = ashpd::AsyncRequestProxy::new_for_path(&connection, request_handle.as_str())?;

    let (sender, receiver) = futures::channel::oneshot::channel();

    let sender = std::cell::Cell::new(Some(sender));

    request
        .connect_response(move |background: Response<Background>| {
            if let Some(m) = sender.replace(None) {
                let _result = m.send(background);
            }
            Box::pin(async { Ok(()) })
        })
        .await?;
```

src/request.rs
```diff
@@ -1,20 +1,108 @@
...
-/// A typical response returned by the `on_response` signal of a `RequestProxy`.
+/// A typical response returned by the `connect_response` signal of a `RequestProxy`.
 ///
 /// [`RequestProxy`]: ./struct.RequestProxy.html
-pub type Response<T> = std::result::Result<T, ResponseError>;
+#[derive(Debug)]
+pub enum Response<T> {
+    Ok(T),
+    Err(ResponseError),
+}
```
sophie-h, line 18: is this necessary?

bilelmoussaoui: Yes, because the responses returned by every request is in the form of (ResponseType, HashMap<String, Value>>), and it's the only sane way to transform it to something that can work across different possible responses

bilelmoussaoui:
> I still think there should be a version of `connect_response` that just returns the response. The other thing that feels like boiler plate is `AsyncRequestProxy` but not sure how to avoid that, because `request_background()` is generated.
> 
> ```rust
>     // working code just to store it somewhere
>     let connection = ashpd::zbus::azync::Connection::new_session().await?;
>     let proxy = background::AsyncBackgroundProxy::new(&connection)?;
> 
>     let request_handle = proxy
>         .request_background(
>             ashpd::WindowIdentifier::default(),
>             background::BackgroundOptions::default().reason(&gettext(
>                 "Schedule and continue running backups in background.",
>             )),
>         )
>         .await?;
> 
>     let request = ashpd::AsyncRequestProxy::new_for_path(&connection, request_handle.as_str())?;
> 
>     let (sender, receiver) = futures::channel::oneshot::channel();
> 
>     let sender = std::cell::Cell::new(Some(sender));
> 
>     request
>         .connect_response(move |background: Response<Background>| {
>             if let Some(m) = sender.replace(None) {
>                 let _result = m.send(background);
>             }
>             Box::pin(async { Ok(()) })
>         })
>         .await?;
> ```

Ah interesting, that would simplify the code a lot

bilelmoussaoui:
> I still think there should be a version of connect_response that just returns the response

I definitely agree, I discussed the matter with Zbus's maintainer on Matrix and ended up opening this issue: https://gitlab.freedesktop.org/dbus/zbus/-/issues/142. It should give us the same variant for every signal which will be very useful in general and allow way simpler async code.

>  The other thing that feels like boiler plate is AsyncRequestProxy but not sure how to avoid that

The idea in the future is for zbus to be able to return a Proxy instead of an ObjectPath with a macro attribute, until then, we will have to do it this way. See https://gitlab.freedesktop.org/dbus/zbus/-/issues/71

Thanks for the feedback !

bilelmoussaoui:
The latest commit should makes the RequestProxy an implementation detail, all the function are now returning it properly instead of returning an ObjectPath.

src/request.rs
sophie-h: Not sure, but maybe Deref to Result could make this a bit simpler to use?

bilelmoussaoui: > Not sure

The exact issue is that we need to implement `zvariant::Type` & `serde::de::Deserialize` for `Result<T>` which is something you can't do

>  but maybe Deref to Result could make this a bit simpler to use?

Ah, I was looking more into the `Try` trait but it's still unstable.

PR #21, opened 2021-08-25
window_identifier: return an error if from_native is called multiple times

Cogitri:
I havent adjusted the other docs yet since I first wanted to see if this approach is OK.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -37,6 +37,8 @@ pub enum Error {
     Zbus(zbus::fdo::Error),
     /// A signal returned no response.
     NoResponse,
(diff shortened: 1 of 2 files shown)


bilelmoussaoui:
Would be nice to return the same error for the gtk3 version as well. Also maybe we can make from_native return an error itself instead of silently falling back to the default implementation

Cogitri:
Sure, I’ll implement that later today, thanks for checking this out :)

bilelmoussaoui:
@Cogitri Interesting, maybe we should ask upstream gtk why we can't have the same behaviour in gtk4 as well?

Cogitri:
image

So basically its not in GTK4 because no one implemented it (yet)

bilelmoussaoui:
Worked around it on ashpd side

PR #25, opened 2021-10-02
Add support for probing camera devices on socket file descriptors

### LINKED ISSUE #23: `--filesystem=xdg-run/pipewire-0` reduces security
https://github.com/bilelmoussaoui/ashpd/blob/e082795083a86c78df5196c15759e34d742fcd8d/src/desktop/camera.rs#L121-L122
The portal doesn't return a node ID, but you can use the file descriptor returned by the portal to iterate the camera node IDs, making exposing `xdg-run/pipewire-0` unneccesary. The portal sets up security restrictions on the socket before returning the socket to the application, but always ensures that the application can iterate and access camera devices.
`--filesystem=xdg-run/pipewire-0` is insecure as it allows the application to effectively bypass the permission prompt.

msizanoen1:
Close #23


bilelmoussaoui:
This also needs to drop the permission from the flatpak manifest of the demo.

Note the build will fail currently but that's caused by a new zbus beta.

bilelmoussaoui:
Would you mind rebasing your PR? The CI should pass correctly now. Thank you

Cargo.toml
```diff
@@ -39,3 +39,4 @@ zvariant = {version = "2.7", features = ["enumflags2"], default-features = false
 zvariant_derive = "2.7"
 futures = "0.3"
 tracing = "0.1"
+libc = "0.2.94"
```
bilelmoussaoui, line 42: The dependency should be enabled only when you use the pipewire feature.

A6GibKm:
Thanks a lot for this MR! I am looking forward to correctly get those ids.

bilelmoussaoui:
Any ideas why I might be getting a `pw::Error::CreationFailed` on `pipewire_node_id` on ashpd-demo?

msizanoen1:
`libc::fcntl(F_DUPFD_CLOEXEC)` returns `-1` when the file descriptor does not exist.

msizanoen1:
#26

PR #26, opened 2021-10-06
Use OwnedFd instead of Fd for returned file descriptors from zbus

msizanoen1:
`zbus` now closes file descriptors when the message is destroyed, unless if you use `OwnedFd`.


bilelmoussaoui:
thanks!

PR #34, opened 2021-12-04
Fix build w/ latest gtk3-rs & gtk4-rs master

Cogitri:


src/desktop/screenshot.rs
```diff
@@ -139,16 +139,16 @@ impl Color {
...
-        gdk3::RGBA::new(color.red(), color.green(), color.blue(), 1.0)
+        gtk3::gdk::RGBA::new(color.red(), color.green(), color.blue(), 1.0)
     }
 }
 
 #[cfg(feature = "feature_gtk4")]
 impl From<Color> for gtk4::gdk::RGBA {
     fn from(color: Color) -> Self {
-        gtk4::gdk::RGBABuilder::new()
+        gtk4::gdk::builders::RGBABuilder::new()
```
bilelmoussaoui, line 151: Would prefer calling RGBA::builder directly

Cogitri: Ah didn't realise we had that, fixed.

PR #41, opened 2022-01-18
window identifier: provide a way API around raw-window-handle

### LINKED ISSUE #40: conversion from WindowIdentifier to raw_window_handle::RawWindowHandle
[raw_window_handle](https://lib.rs/crates/raw-window-handle) is a cross platform library abstracting over window handles for various windowing systems, including Wayland and X11. For cross platform code using ashpd, for example [rfd](https://github.com/PolyMeilex/rfd/pull/41/files#r786441219), it would be helpful for ashpd to provide an API to convert its WindowIdentifier type to raw_window_handle's RawWindowHandle type. Alternatively, perhaps ashpd could switch to using RawWindowHandle in its own API.

bilelmoussaoui:
Fixes #40 

Still needs an API to get a Window Identifier from a RawWindowHandle.  Also needs a new release of gdk3wayland to expose the ffi crate

### DIFF
--- a/src/window_identifier.rs
+++ b/src/window_identifier.rs
@@ -17,6 +17,9 @@ use gtk3::{
     glib::{self, clone},
     prelude::*,
 };
+
+#[cfg(feature = "raw_handle")]
+use raw_window_handle::{RawWindowHandle, WaylandHandle, XlibHandle};
 /// Most portals interact with the user by showing dialogs.
 /// These dialogs should generally be placed on top of the application window
 /// that triggered them. To arrange this, the compositor needs to know about the
@@ -70,6 +73,12 @@ use gtk3::{
 ///
 /// ## Other Toolkits
 ///
+/// If you have access to `RawWindowHandle` you can convert it to a [`WindowIdentifier`] with
+/// ```rust, ignore
+///     let handle = RawWindowHandle::Xlib(XlibHandle::empty());
+///     let identifier = WindowIdentifier::from_raw_handle(handle).await;///
+/// ```
+///
 /// In case you don't have access to a WindowIdentifier:
 ///
 /// ```rust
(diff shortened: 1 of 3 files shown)


Be-ing:
Thanks for implementing this!

src/window_identifier.rs
```diff
@@ -241,11 +246,157 @@ impl WindowIdentifier {
...
+    ///
+    /// # Panics
+    ///
+    /// If the passed RawWindowHandle is not of `WaylandHandle` or `XlibHandle` type. As other platforms handles
+    /// are not supported by the portals so far.
+    pub async fn form_raw_handle(handle: RawWindowHandle) -> Self {
+        use raw_window_handle::RawWindowHandle::{Wayland, Xlib};
+        match handle {
+            // TODO: figure out how to make use of the wl_display to export a handle with xdg-foreign protocol
+            Wayland(_wl_handle) => Self::default(),
```
bilelmoussaoui, line 266: cc @Be-ing, this is sub-optimal but I have no idea how to do this using the wayland-client crate. It seems like we could make use of https://smithay.github.io/wayland-rs/wayland_protocols/unstable/xdg_foreign/v2/client/zxdg_exporter_v2/struct.ZxdgExporterV2.html but then I can't figure out how to get an instance of `ConnectionHandle`

bilelmoussaoui: I don't intend to spend any more time on implementing this part, so someone else will have to deal with that.

PolyMeilex: This will be problematic, especially after the migration of the wayland-rs ecosystem to 0.30, as ConnectionHandle is something that user owns, so the only way is to pass it down.

@vberger Sorry for a ping, but do you perhaps have any thoughts on this? For context, we are trying to export a toplevel for use in XDG Portal, but raw-window-handle only provides two raw pointers to `surface` and `display`, so we have no way to create resources or sent any requests to the server. My initial idea is to just pass conn to it, like we do in Smithay and sctl, but having a Wayland specific function kinda defeats the purpose or `raw_window_handle`

bilelmoussaoui: @PolyMeilex the code already makes use of 0.30  btw

PolyMeilex: Yep, I meant that ecosystem is not ported yet, like `sctl` and `winit`

PR #45, opened 2022-01-26
fix typo in function name (form -> from)

Be-ing:


src/window_identifier.rs
```diff
@@ -285,7 +285,7 @@ impl WindowIdentifier {
     /// Create an instance of [`WindowIdentifier`] from a [`RawWindowHandle`](raw_window_handle::RawWindowHandle).
     ///
     /// The constructor returns a valid handle under both Wayland & x11.
-    pub fn form_raw_handle(handle: &RawWindowHandle) -> Self {
```
bilelmoussaoui, line 288: Oh no

bilelmoussaoui:
I will merge this once the CI passes and do 0.2.1 release with that

bilelmoussaoui:
0.2.1 should be out

Be-ing:
Thanks. This is technically a breaking change, but I am quite doubtful anyone is actually depending on this typo considering it was released just yesterday and anyone trying to use it would have noticed the typo.

bilelmoussaoui:
Yes I'm aware that is a breaking change but it is a part of the API that is very unlikely to be used by the most common users of ashpd which are applications that is why I didn't want to bump to 0.3 just for that :)

bilelmoussaoui:
That and also being behind a feature flag

PR #46, opened 2022-01-27
window identifier: store the wayland handle in the object data

### LINKED ISSUE #20: WindowIdentifier::from_native() may only be called once for Wayland GTK windows
Hello,

it seems like WindowIndentifier::from_native() can only be called once, otherwise the following GCritical is emitted:

```
(dev.Cogitri.Health:21): Gdk-CRITICAL **: 12:22:08.835: gdk_wayland_toplevel_export_handle: assertion '!impl->display_server.xdg_exported' failed
```

This GCritical origins from `gdk_wayland_toplevel_export_handle`. The documentation for that function says:

```
* gdk_wayland_toplevel_export_handle:
[...]
* It is an error to call this function on a surface that is already
* exported.
```

Not sure what would be the best way to handle this, should we just put a notice in the docs?

bilelmoussaoui:
This way we always return a valid handle even if the WindowIdentifier is constructed multiple times
from different calls

Closes #20 
Supersedes #21 

cc @Cogitri 

This is barely a work around, and it will still fail because the moment the first window identifier is dropped, the handle will no longer be valid because it has been un-exported after that. I think we can work around that issue as well, by storing a refcount and only unexport if that refcount is equal to 1. Edit: fixed

Before I do the rest of that, I was wondering what do you think of this approach? If all sounds good to you, I will implement the rest, test it on ashpd-demo and push it as part of 0.2.2 release.

### DIFF
--- a/src/window_identifier.rs
+++ b/src/window_identifier.rs
@@ -135,6 +135,9 @@ pub enum WindowIdentifier {
     Other(String),
 }
 
+#[cfg(any(feature = "feature_gtk4", feature = "feature_gtk3"))]
+static WINDOW_HANDLE_KEY: &str = "ashpd-wayland-window-handle";
+
 unsafe impl Send for WindowIdentifier {}
 unsafe impl Sync for WindowIdentifier {}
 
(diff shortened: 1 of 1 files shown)


bilelmoussaoui:
As expected, it works nicely!

PR #50, opened 2022-03-07
Relax bounds for AsRef<Path>.

jneem:
I think the `Type` and `Serialize` parts of the `AsRef<Path> + Type + Serialize` weren't needed, because the path just gets turned into a byte slice anyway (which implements `Type` and `Serialize`).


bilelmoussaoui (approved):

bilelmoussaoui:
Thanks!

bilelmoussaoui:
I have released 0.2.6 with the changes on this PR in case you need it.

PR #51, opened 2022-03-17
Better support for WindowIdentifiers on x11.

jneem:
This adds support for XCB in `from_raw_handle`, and it also allows
for the creation of a WindowIdentifier on X11 without bringing in
the extra dependencies required for the `raw_handle` feature.


bilelmoussaoui:
Nice, thanks

PR #57, opened 2022-03-22
Desktop: Add Icon type

### LINKED ISSUE #56: Provide a wrapper for GIcon
The notification/dynamic launcher portals expect a serialized GIcon, currently there is no easy way for apps to provide that if they are not using gtk and even then you can't easily pass the result of a `g_icon_serialize` as a `zvariant::Value`. 

The idea is to provide an enum that implements Serialize/Deserialize in a similar way how `g_icon_serialize`/ `g_icon_deserialize` works

bilelmoussaoui:
it is basically a rewrite of the GIcon serialize implementation
that Notification/Dynamic Launcher expects

Fixes #56

### DIFF
--- a/ashpd-demo/src/portals/desktop/email.rs
+++ b/ashpd-demo/src/portals/desktop/email.rs
@@ -76,19 +76,19 @@ impl EmailPage {
         let root = self.native().unwrap();
 
         let mut email = Email::new();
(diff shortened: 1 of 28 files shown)


bilelmoussaoui:
This was merged already

PR #60, opened 2022-04-18
Update pipewire to 0.5

Newbytee:
Should let ashpd build on 32-bit architectures with glibc and gets it closer to building on 32-bit architectures with musl.


bilelmoussaoui:
Thanks!

PR #67, opened 2022-06-04
Add option to use async-std or tokio

A6GibKm:
As it is, it requires either async-std or tokio which might be undesirable.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,7 +16,7 @@ feature_gtk3 = ["gdk3x11", "gdk3wayland", "gtk3"]
 feature_gtk4 = ["gdk4x11", "gdk4wayland", "gtk4"]
 feature_pipewire = ["pw", "libc"]
 log = ["tracing"]
-tokio_runtime = ["zbus/tokio", "tokio"]
+tokio = ["zbus/tokio", "dep:tokio"]
 raw_handle = ["raw-window-handle", "wayland"]
 wayland = ["wayland-client", "wayland-protocols", "wayland-backend"]
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,12 +11,12 @@ repository = "https://github.com/bilelmoussaoui/ashpd"
 version = "0.4.0"
 
 [features]
-default_features = []
+default = ["async-std"]
 feature_gtk3 = ["gdk3x11", "gdk3wayland", "gtk3"]
 feature_gtk4 = ["gdk4x11", "gdk4wayland", "gtk4"]
 feature_pipewire = ["pw", "libc"]
 log = ["tracing"]
-tokio_runtime = ["zbus/tokio"]
+tokio_runtime = ["zbus/tokio", "tokio"]
 raw_handle = ["raw-window-handle", "wayland"]
 wayland = ["wayland-client", "wayland-protocols", "wayland-backend"]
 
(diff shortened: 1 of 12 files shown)


Cargo.toml
```diff
@@ -11,12 +11,12 @@ repository = "https://github.com/bilelmoussaoui/ashpd"
 version = "0.4.0"
 
 [features]
-default_features = []
+default_features = ["async-std"]
```
bilelmoussaoui, line 14:
```suggestion
default = ["async-std"]
```

Cargo.toml
```diff
@@ -42,6 +42,8 @@ raw-window-handle = {version = "0.4", optional = true}
 wayland-client = {version = "0.30.0-alpha2", optional = true}
 wayland-protocols = {version = "0.30.0-alpha2", optional = true, features = ["unstable_protocols", "client"]}
 wayland-backend = {version = "0.1.0-alpha2", optional = true, features = ["client_system"]}
+async-std = {version = "1.11", optional = true}
+tokio = {version = "1.17", features = ["fs", "io-util"], optional = true}
```
bilelmoussaoui, line 46:
```suggestion
tokio = {version = "1.17", features = ["fs", "io-util"], default_features = false, optional = true}
```

src/desktop/secret.rs
```diff
@@ -19,8 +19,12 @@
 //! }
 //! ```
 
+#[cfg(not(feature = "async-std"))]
```
bilelmoussaoui, line 22: tokio doesn't have an equivalent for this?

A6GibKm: it does, yes.

Cargo.toml
```diff
@@ -11,12 +11,12 @@ repository = "https://github.com/bilelmoussaoui/ashpd"
 version = "0.4.0"
 
 [features]
-default_features = []
+default = ["async-std"]
```
bilelmoussaoui, line 14: Make sure to update https://github.com/bilelmoussaoui/ashpd#optional-features

Cargo.toml
```diff
@@ -16,7 +16,7 @@ feature_gtk3 = ["gdk3x11", "gdk3wayland", "gtk3"]
 feature_gtk4 = ["gdk4x11", "gdk4wayland", "gtk4"]
 feature_pipewire = ["pw", "libc"]
 log = ["tracing"]
-tokio_runtime = ["zbus/tokio", "tokio"]
+tokio = ["zbus/tokio", "dep:tokio"]
```
bilelmoussaoui, line 19: You also need to do changes in code where you were using feature = "tokio_runtime". 

If you are really motivated to do that, please drop the log feature -> make it use tracing as it is an optional dependency already. Rename feature_gtk4/feature_gtk3/feature_pipewire to gtk4/gtk3/pipewire

src/desktop/secret.rs
```diff
@@ -19,8 +19,12 @@
 //! }
 //! ```
 
-use std::{io::Read, os::unix::prelude::AsRawFd};
+use std::os::unix::prelude::AsRawFd;
 
+#[cfg(feature = "async-std")]
+use async_std::{os::unix::net::UnixStream, prelude::*};
+#[cfg(all(feature = "tokio_runtime", not(feature = "async-std")))]
```
bilelmoussaoui, line 26: i would keep feature=tokio_runtime without the async-std check & add a compile_error on lib.rs similar to when you enable both gtk3/gtk4

A6GibKm: Done. Added it to oo7 PR too.

Cargo.toml
A6GibKm: Done, but I wonder if it is a good idea to rename the others.... well it is in its own commit which can be removed.

PR #70, opened 2022-06-04
WIP: Add activation token support

A6GibKm:
TODO:

 - [ ] Do GTK 3 implementation
 - [ ] Do GTK 4 implementation
 - [x] Do wayland implementation
 - [x] Implement it for the portals that make sense
   -  [x] OpenUri (OpenURI, OpenFile, OpenDirectory)
   -  [x] DynamicLauncher (Launch)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,7 +29,7 @@ gdk3x11 = {package = "gdkx11", version = "0.16", optional = true}
 gdk3wayland = {package = "gdkwayland", version = "0.16", optional = true}
 gtk3 = {package = "gtk", version = "0.16", optional = true}
 
-gdk4wayland = {package = "gdk4-wayland", version = "0.5", optional = true}
+gdk4wayland = {package = "gdk4-wayland", version = "0.5", optional = true, features = ["wayland_crate"]}
 gdk4x11 = {package = "gdk4-x11", version = "0.5", optional = true}
 gtk4 = {version = "0.5", optional = true}
 
@@ -44,7 +44,7 @@ tracing = {version = "0.1", optional = true}
 libc = {version = "0.2", optional = true}
 raw-window-handle = {version = "0.5", optional = true}
 wayland-client = {version = "0.30", optional = true}
-wayland-protocols = {version = "0.30", optional = true, features = ["unstable", "client"]}
+wayland-protocols = {version = "0.30", optional = true, features = ["unstable", "client", "staging"]}
 wayland-backend = {version = "0.1", optional = true, features = ["client_system"]}
 async-std = {version = "1.12", optional = true}
 tokio = {version = "1.21", features = ["fs", "io-util"], optional = true, default-features = false}
--- a/src/activation_token/gtk3.rs
+++ b/src/activation_token/gtk3.rs
@@ -1,0 +1,17 @@
+use gtk3::prelude::*;
+use gtk3::{gdk, glib};
+
+#[derive(Debug)]
+pub struct Gtk3ActivationToken {
+    pub(crate) token: String,
+}
+
+impl Gtk3ActivationToken {
+    pub fn from_window(window: &impl glib::IsA<gdk::Window>) -> Option<Self> {
+        let display = window.as_ref().display();
+        match display.backend() {
+            gdk::Backend::Wayland => todo!(),
+            _ => None,
+        }
+    }
+}
--- a/src/activation_token/gtk4.rs
+++ b/src/activation_token/gtk4.rs
@@ -1,0 +1,40 @@
+#[cfg(feature = "wayland")]
+use super::wayland::WaylandActivationToken;
+use gdk4wayland::prelude::WaylandSurfaceExtManual;
+use gtk4::{gdk, glib, prelude::*};
+
+#[derive(Debug)]
+pub struct Gtk4ActivationToken {
+    pub(crate) wl_token: WaylandActivationToken,
+}
+
+#[cfg(all(feature = "gtk4_wayland", feature = "wayland"))]
+impl Gtk4ActivationToken {
+    pub async fn from_native<N: glib::IsA<gtk4::Native>>(app_id: &str, native: &N) -> Option<Self> {
+        let surface = native.surface();
+        match surface.display().backend() {
+            gdk::Backend::Wayland => {
+                let surface = surface
+                    .downcast_ref::<gdk4wayland::WaylandSurface>()
+                    .unwrap();
+                if let Some(wl_surface) = surface.wl_surface() {
+                    let wl_token = WaylandActivationToken::from_surface(app_id, &wl_surface)
+                        .await
+                        .unwrap();
+
+                    Some(Self { wl_token })
+                } else {
+                    None
+                }
+            }
+            _ => None,
+        }
+    }
+}
+
+#[cfg(feature = "wayland")]
+impl From<WaylandActivationToken> for Gtk4ActivationToken {
+    fn from(wl_token: WaylandActivationToken) -> Self {
+        Self { wl_token }
+    }
+}
--- a/src/activation_token/mod.rs
+++ b/src/activation_token/mod.rs
@@ -1,0 +1,149 @@
+#[cfg(all(feature = "gtk3", feature = "wayland"))]
+mod gtk3;
+#[cfg(all(feature = "gtk3", feature = "wayland"))]
+pub use self::gtk3::Gtk3ActivationToken;
+
+#[cfg(feature = "gtk4_wayland")]
+mod gtk4;
+#[cfg(feature = "gtk4_wayland")]
+pub use self::gtk4::Gtk4ActivationToken;
+
+#[cfg(any(feature = "wayland"))]
+mod wayland;
+#[cfg(feature = "wayland")]
+pub use wayland::WaylandActivationToken;
+
+use serde::{ser::Serializer, Serialize};
+use zbus::zvariant::Type;
+
+// TODO
+/// See https://wayland.app/protocols/xdg-activation-v1
+#[derive(Debug, Type)]
+#[zvariant(signature = "s")]
+pub enum ActivationToken {
+    #[cfg(feature = "wayland")]
+    #[doc(hidden)]
+    Wayland(WaylandActivationToken),
+    #[cfg(feature = "gtk4_wayland")]
+    #[doc(hidden)]
+    Gtk4(Gtk4ActivationToken),
+    #[cfg(all(feature = "gtk3", feature = "wayland"))]
+    #[doc(hidden)]
+    Gtk3(Gtk3ActivationToken),
+    #[doc(hidden)]
+    Raw(String),
+    #[doc(hidden)]
+    None,
+}
+
+impl Serialize for ActivationToken {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        serializer.serialize_str(self.as_str())
+    }
+}
+
+impl Default for ActivationToken {
+    fn default() -> Self {
+        Self::None
+    }
+}
+
+impl ActivationToken {
+    #[cfg(feature = "wayland")]
+    /// Create an instance of [`ActivationToken`] from a Wayland surface and the
+    /// application's id.
+    pub async fn from_wayland_surface(
+        app_id: &str,
+        surface: &wayland_client::protocol::wl_surface::WlSurface,
+    ) -> Self {
+        if let Some(token) = WaylandActivationToken::from_surface(app_id, surface).await {
+            Self::Wayland(token)
+        } else {
+            Self::default()
+        }
+    }
+
+    #[cfg(feature = "wayland")]
+    /// Create an instance of [`ActivationToken`] from a raw Wayland surface and
+    /// the application's id.
+    ///
+    /// # Safety
+    ///
+    /// Both pointers have to be valid surface and display pointers. You must
+    /// ensure the `display_ptr` lives longer than the returned
+    /// `ActivationToken`.
+    pub async unsafe fn from_wayland_raw(
+        app_id: &str,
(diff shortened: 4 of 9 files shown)


src/window_identifier/wayland.rs
```diff
@@ -74,6 +80,24 @@ impl wayland_client::Dispatch<ZxdgExportedV2, ()> for ExportedWaylandHandle {
     }
 }
 
+impl wayland_client::Dispatch<XdgActivationTokenV1, ()> for super::ActivationToken {
```
bilelmoussaoui, line 83: this should live in a new file named `activation_token.rs`

src/window_identifier/wayland.rs
```diff
@@ -124,3 +148,18 @@ fn wayland_handle_unexport(exported: &ZxdgExportedV2) -> Result<(), Box<dyn std:
 
     Ok(())
 }
+
+pub(super) fn activation_token_from_surface(
```
bilelmoussaoui, line 152: Same

src/window_identifier/mod.rs
```diff
@@ -254,6 +254,23 @@ impl WindowIdentifier {
     }
 }
 
+#[derive(Default, Debug)]
```
bilelmoussaoui, line 257: Same as the rest

src/window_identifier/gtk4.rs
```diff
@@ -140,3 +117,23 @@ impl Drop for Gtk4WindowIdentifier {
         }
     }
 }
+
+pub(super) fn activation_token_from_native(
```
bilelmoussaoui, line 121: Move this to the activation token type?

src/window_identifier/gtk4.rs
```diff
@@ -26,35 +21,17 @@ impl Gtk4WindowIdentifier {
         let surface = native.surface();
         match surface.display().backend() {
             Backend::Wayland => {
-                let top_level = surface
+                use gdk4wayland::prelude::WaylandSurfaceExtManual;
```
bilelmoussaoui, line 24: You need to modify the drop implementation

src/desktop/open_uri.rs
```diff
@@ -303,7 +310,9 @@ pub async fn open_file(
 ) -> Result<(), Error> {
     let connection = zbus::Connection::session().await?;
     let proxy = OpenURIProxy::new(&connection).await?;
-    proxy.open_file(identifier, file, writeable, ask).await?;
+    proxy
```
bilelmoussaoui, line 313: This function could be modified to give it the activation token as param

src/desktop/open_uri.rs
```diff
@@ -233,8 +236,12 @@ impl<'a> OpenURIProxy<'a> {
         file: &impl AsRawFd,
         writeable: bool,
         ask: bool,
+        activation_token: Option<ActivationToken>,
     ) -> Result<(), Error> {
-        let options = OpenFileOptions::default().ask(ask).writeable(writeable);
+        let mut options = OpenFileOptions::default().ask(ask).writeable(writeable);
+        if let Some(token) = activation_token {
```
bilelmoussaoui, line 242: Make the option take ActivationToken instead of a string & impelement Serialize on it

bilelmoussaoui: Are there other portals that take this option and we don't handle currently? needs to be checked

bilelmoussaoui: Seems not, there is DynamicLauncher portal if you want to do it. It needs a new struct to define the options for it though as it doesn't have one for now

src/window_identifier/gtk4.rs
bilelmoussaoui: Also commit: 
```
WindowIdentifier/gtk4: Use wayland crate for exporting the surface

As we already need wayland for the activation token support,see next commit.
This will also avoid us having custom logic that is needed for keeping an internal ref counted token
because gtk4 doesn't support calling export_handle on an already exported surface.
```

bilelmoussaoui: Also this function no longer needs to be async, yay!

src/window_identifier/mod.rs
```diff
@@ -254,6 +254,23 @@ impl WindowIdentifier {
...
+pub struct ActivationToken(pub(crate) String);
```
bilelmoussaoui, line 258: the new type needs to be exported in lib.rs & could use some very basic docs + a link the wayland protocol.

src/desktop/open_uri.rs
A6GibKm: Do we really need a struct? I used a simple hashmap here as it was already the case.

bilelmoussaoui: The struct makes it super easy to enforce the type system instead of a random map somewhere

src/activation_token/gtk3.rs
```diff
@@ -0,0 +1,60 @@
...
+                    );
+
+                    ObjectId::from_ptr(&WL_SURFACE_INTERFACE, ptr as *mut _).unwrap()
+                };
+                let wl_surface = WlSurface::from_id(&cnx, surface_id).unwrap();
+                let wl = WaylandActivationToken::from_surface(&wl_surface).ok()?;
+
+                Some(Self::from(wl))
+            }
+            // TODO Can this be implemented for X11?
```
bilelmoussaoui, line 40: Nop it is not

src/activation_token/gtk4.rs
```diff
@@ -0,0 +1,48 @@
...
+                    .surface()
+                    .downcast_ref::<gdk4wayland::WaylandToplevel>()
+                    .unwrap()
+                    .wl_surface();
+
+                let wl = WaylandActivationToken::from_surface(&wl_surface).ok()?;
+
+                Some(Self::from(wl))
+            }
+            // TODO Can this be implemented for X11?
```
bilelmoussaoui, line 28: Same here

src/activation_token/gtk3.rs
```diff
@@ -0,0 +1,60 @@
...
+                    .unwrap();
+                let surface_id = unsafe {
+                    let ptr = gdk3wayland::ffi::gdk_wayland_window_get_wl_surface(
+                        wayland_window.to_glib_none().0,
+                    );
+
+                    ObjectId::from_ptr(&WL_SURFACE_INTERFACE, ptr as *mut _).unwrap()
+                };
+                let wl_surface = WlSurface::from_id(&cnx, surface_id).unwrap();
+                let wl = WaylandActivationToken::from_surface(&wl_surface).ok()?;
```
bilelmoussaoui, line 36: Could use a custom constructor that takes a pointer and use it here & and in the gtk4 variant maybe?

src/activation_token/mod.rs
```diff
@@ -0,0 +1,82 @@
...
+    {
+        serializer.serialize_str(self.as_str())
+    }
+}
+
+impl ActivationToken {
+    #[cfg(feature = "wayland")]
+    pub fn from_surface(
+        surface: &wayland_client::protocol::wl_surface::WlSurface,
+    ) -> Result<Self, Box<dyn std::error::Error>> {
```
bilelmoussaoui, line 51: Why this can fail and not the rest?

src/activation_token/wayland.rs
```diff
@@ -0,0 +1,59 @@
...
+        if let Some(wl_token) = self.inner.take() {
+            wl_token.destroy();
+        }
+    }
+}
+
+impl WaylandActivationToken {
+    pub fn from_surface(surface: &WlSurface) -> Result<Self, Box<dyn std::error::Error>> {
+        let cnx = wayland_client::Connection::connect_to_env().unwrap();
+        let wl_activation = XdgActivationV1::from_id(&cnx, surface.id()).unwrap();
```
PolyMeilex, line 25: This look like it will always panic, ID of a WlSurface will never be equal to ID of XdgActivation object.

Cargo.toml
```diff
@@ -29,7 +29,7 @@ gdk3x11 = {package = "gdkx11", version = "0.16", optional = true}
 gdk3wayland = {package = "gdkwayland", version = "0.16", optional = true}
 gtk3 = {package = "gtk", version = "0.16", optional = true}
 
-gdk4wayland = {package = "gdk4-wayland", version = "0.5", optional = true}
+gdk4wayland = {package = "gdk4-wayland", version = "0.5", optional = true, features = ["wayland_crate"]}
```
A6GibKm, line 32: I don't think this is correct, the feature should be enabled only if `gtk4_wayland` is enabled.

src/desktop/open_uri.rs
```diff
@@ -57,14 +57,14 @@ use zbus::zvariant::{Fd, SerializeDict, Type};
...
-    Error, WindowIdentifier,
+    ActivationToken, Error, WindowIdentifier,
 };
 
 #[derive(SerializeDict, Type, Debug, Default)]
 #[zvariant(signature = "dict")]
 struct OpenDirOptions {
     handle_token: HandleToken,
-    activation_token: Option<String>,
+    activation_token: ActivationToken,
```
bilelmoussaoui, line 67: It is optional so why make it no longer the case?

Cargo.toml
bilelmoussaoui: not sure what you mean here?

bilelmoussaoui: that looks correct to me

src/activation_token/mod.rs
```diff
@@ -0,0 +1,119 @@
...
+        Self::None
+    }
+}
+
+impl ActivationToken {
+    #[cfg(feature = "wayland")]
+    /// Create an instance of [`ActivationToken`] from a Wayland surface and the
+    /// application's id.
+    pub async fn from_wayland_surface(
+        app_id: &str,
```
bilelmoussaoui, line 60: We already have plenty of places where we use the `app_id` so maybe we should have a type for it that wraps a `String` or something like that.

src/desktop/open_uri.rs
A6GibKm: There is ActivationToken::None which is the default so wrapping in option seems redundant.

src/activation_token/mod.rs
A6GibKm: I think its way more convenient for the user to just write their app_id in a &str, ideally it would be preferable to somehow get it from the wayland toplevel or gtkwindow..

bilelmoussaoui: Sure, you could make those functions take a `impl Into<AppID>` and `impl From<&str> for AppID` but as it is something that can be done in a separate PR, I will handle it then you can just rebase.

src/activation_token/mod.rs
```diff
@@ -0,0 +1,149 @@
...
+    pub(crate) fn as_str(&self) -> &str {
+        match self {
+            #[cfg(feature = "wayland")]
+            Self::Wayland(activation_token) => activation_token.token.as_str(),
+            #[cfg(feature = "gtk4_wayland")]
+            Self::Gtk4(activation_token) => activation_token.wl_token.token.as_str(),
+            #[cfg(all(feature = "gtk3", feature = "wayland"))]
+            Self::Gtk3(activation_token) => activation_token.token.as_str(),
+            Self::Raw(string) => string.as_str(),
+            Self::None => "",
```
bilelmoussaoui, line 133: What is the difference between None and Raw? You can make the default impl use `Raw("")`? :)

src/desktop/dynamic_launcher.rs
```diff
@@ -248,9 +248,16 @@ impl<'a> DynamicLauncherProxy<'a> {
     /// See also [`Launch`](https://flatpak.github.io/xdg-desktop-portal/index.html#gdbus-method-org-freedesktop-portal-DynamicLauncher.Launch).
     #[doc(alias = "Launch")]
     #[doc(alias = "xdp_portal_dynamic_launcher_launch")]
-    pub async fn launch(&self, desktop_file_id: &str) -> Result<(), Error> {
-        // TODO: handle activation_token
-        let options: HashMap<&str, zvariant::Value<'_>> = HashMap::new();
+    pub async fn launch(
+        &self,
+        desktop_file_id: &str,
+        activation_token: ActivationToken,
```
bilelmoussaoui, line 254: Should definitely be `impl Into<Option<ActivationToken>>`

src/desktop/open_uri.rs
bilelmoussaoui: Sure, but it is simpler to take `Impl<Into<Option<T>>` that way you don't have to handle both cases in your app side :) see all the other options.

src/desktop/open_uri.rs
```diff
@@ -171,6 +171,15 @@ impl OpenFileRequest {
         self
     }
 
+    #[must_use]
+    /// Sets the activation token for the application.
+    pub fn activation_token(mut self, activation_token: ActivationToken) -> Self {
+        if activation_token.is_some() {
```
bilelmoussaoui, line 177: that is why I think this should definitely be optional instead of these workarounds

src/activation_token/mod.rs
```diff
@@ -0,0 +1,149 @@
...
+pub use wayland::WaylandActivationToken;
+
+use serde::{ser::Serializer, Serialize};
+use zbus::zvariant::Type;
+
+// TODO
+/// See https://wayland.app/protocols/xdg-activation-v1
+#[derive(Debug, Type)]
+#[zvariant(signature = "s")]
+pub enum ActivationToken {
```
bilelmoussaoui, line 23: We should add a TODO here to implement a X11 only thing.

bilelmoussaoui:
It seems like the only useful bit left in this PR is the wayland implementation, is that something you plan to push forward?

A6GibKm:
Probably, but not at the moment.

bilelmoussaoui:
I will push it forward myself then, thanks!

bilelmoussaoui:
Most of the interesting bits from this PR have now landed, might need improvements but could be done along the way

Thank you!

PR #75, opened 2022-06-08
Drop the Proxy suffix

bilelmoussaoui:
This also renames some types to simplify things.
Currently only EmailProxy have not been renamed but
the plan for this one is to convert it to a builder pattern like type
and keep the Proxy as an internal implementation API


bilelmoussaoui:
The more I think about this, I wonder if we can just make all the proxy types internals and have fancy & simple to use wrappers on top.

PR #76, opened 2022-06-09
Fix wayland window identifier

A6GibKm:


src/window_identifier/wayland.rs
```diff
@@ -1,40 +1,57 @@
...
-        match wayland_handle_export(surface_ptr) {
-            Ok((exported, handle)) => Some(Self {
-                exported,
-                type_: WindowIdentifierType::Wayland(handle),
-            }),
-            _ => None,
-        }
+    pub async fn from_raw(
+        surface_ptr: *mut std::ffi::c_void,
+        display_ptr: *mut std::ffi::c_void,
```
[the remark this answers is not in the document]
A6GibKm: Done.

src/window_identifier/wayland.rs
```diff
@@ -1,40 +1,57 @@
...
 
-    pub fn from_raw(surface_ptr: *mut std::ffi::c_void) -> Option<Self> {
-        match wayland_handle_export(surface_ptr) {
-            Ok((exported, handle)) => Some(Self {
-                exported,
-                type_: WindowIdentifierType::Wayland(handle),
-            }),
-            _ => None,
-        }
+    pub async fn from_raw(
```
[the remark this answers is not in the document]
A6GibKm: Done.

src/window_identifier/wayland.rs
```diff
@@ -66,61 +108,151 @@ impl wayland_client::Dispatch<ZxdgExportedV2, ()> for ExportedWaylandHandle {
...
-fn wayland_handle_unexport(exported: &ZxdgExportedV2) -> Result<(), Box<dyn std::error::Error>> {
-    exported.destroy();
 
-    Ok(())
+async fn wayland_export_handle(surface: &WlSurface) -> Option<WaylandWindowIdentifier> {
+    let (sender, receiver) =
+        futures::channel::oneshot::channel::<Option<WaylandWindowIdentifier>>();
+
+    // Can we do better?
+    let surface = surface.clone();
```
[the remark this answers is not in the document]
A6GibKm: Thanks for the info

src/window_identifier/wayland.rs
```diff
@@ -1,40 +1,79 @@
...
+        let surface = WlSurface::from_id(&conn, obj_id).ok()?;
+
+        Self::new_inner(conn, &surface).await
+    }
+
+    async fn new_inner(conn: wayland_client::Connection, surface: &WlSurface) -> Option<Self> {
+        let (sender, receiver) =
+            futures::channel::oneshot::channel::<Option<WaylandWindowIdentifier>>();
+
+        // Cheap clone, protocol objects are essentially smart pointers
```
bilelmoussaoui, line 65:
```suggestion
        // Cheap clone, protocol objects are essentially small pointers
```
iirc

Cargo.toml
```diff
@@ -38,9 +38,9 @@ futures = "0.3"
 tracing = {version = "0.1", optional = true}
 libc = {version = "0.2.94", optional = true}
 raw-window-handle = {version = "0.4", optional = true}
-wayland-client = {version = "0.30.0-beta.3", optional = true}
-wayland-protocols = {version = "0.30.0-beta.3", optional = true, features = ["unstable", "client"]}
-wayland-backend = {version = "0.1.0-beta.3", optional = true, features = ["client_system"]}
+wayland-client = {git = "https://github.com/Smithay/wayland-rs", optional = true}
+wayland-protocols = {git = "https://github.com/Smithay/wayland-rs", optional = true, features = ["unstable", "client"]}
+wayland-backend = {git = "https://github.com/Smithay/wayland-rs", optional = true, features = ["client_system"]}
```
i509VCB, line 43: beta.6 is out with the from_ptr and other functions

src/window_identifier/wayland.rs
```diff
@@ -1,40 +1,57 @@
...
-            _ => None,
-        }
+    pub async fn from_raw(
+        surface_ptr: *mut std::ffi::c_void,
+        display_ptr: *mut std::ffi::c_void,
+    ) -> Option<Self> {
+        let backend = unsafe {
+            wayland_backend::sys::client::Backend::from_foreign_display(display_ptr as *mut _)
+        };
+        let conn = wayland_client::Connection::from_backend(backend);
```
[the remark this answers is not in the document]
i509VCB: I don't think we have much of an option to avoid this.

Plus the `from_foreign_display` will act as a guest backend, so all the handles will still belong to the same underlying display.

src/window_identifier/mod.rs
```diff
@@ -217,18 +217,21 @@ impl WindowIdentifier {
     ///
     /// ## Safety
     ///
-    /// The surface has to be a valid Wayland surface pointer.
-    pub unsafe fn from_wayland_raw(surface_ptr: *mut std::ffi::c_void) -> Self {
-        match WaylandWindowIdentifier::from_raw(surface_ptr) {
+    /// The surface and display have to be valid Wayland surface pointers.
+    pub async unsafe fn from_wayland_raw(
```
i509VCB, line 221: Add the following safety text:

```rust
/// # Safety
///
/// You must ensure the `display_ptr` lives longer than the returned `WindowIdentifier`.
```

A6GibKm: Used
```rust
    /// Create an instance of [`WindowIdentifier`] from a Wayland surface.
    ///
    /// # Safety
    ///
    /// Both pointers have to be valid surface and display pointers. You must
    /// ensure the `display_ptr` lives longer than the returned
    /// `WindowIdentifier`.
```

src/window_identifier/wayland.rs
A6GibKm: Smart was the word used at https://matrix.to/#/#smithay:matrix.org when I asked about it.

PolyMeilex: I think that smart pointer is the correct term here, wayland-rs objects are made of `ObjectId` and ref counted user data, so they essentially act like smart pointer.

src/window_identifier/wayland.rs
```diff
@@ -65,62 +126,135 @@ impl wayland_client::Dispatch<ZxdgExportedV2, ()> for ExportedWaylandHandle {
...
+                    self.exporter_v2 = Some(exporter);
+                }
+                _ => (),
+            }
+        }
+    }
 }
 
-/// A helper to export a wayland handle from a WLSurface
+/// A helper to export a wayland handle from a WlSurface and a WlConnection
```
i509VCB, line 207:
```suggestion
/// A helper to export a wayland handle from a WlSurface and a connection
```

src/window_identifier/mod.rs
```diff
@@ -215,20 +215,25 @@ impl WindowIdentifier {
...
-    /// ## Safety
+    /// # Safety
     ///
-    /// The surface has to be a valid Wayland surface pointer.
-    pub unsafe fn from_wayland_raw(surface_ptr: *mut std::ffi::c_void) -> Self {
-        match WaylandWindowIdentifier::from_raw(surface_ptr) {
+    /// Both pointers have to be valid surface and display pointers. You must
+    /// ensure the `display_ptr` lives longer than the returned
+    /// `WindowIdentifier`.
+    pub async unsafe fn from_wayland_raw(
```
i509VCB, line 223: I'm actually considering whether this function is necessary. Supporting this function introduces a bunch of safety requirements and lifetime questions we need to ask about the connection.

I think what we can do is explain where the user would create the required types from the surface and display pointers (wayland-backend) and then explain what lifetime constraints apply to such handles `WaylandWindowIdentifier` has (must not outlive the connection).

This would also mean the `client_system` feature would not be needed as ashpd would not be responsible for that and would operate on wayland-backend common types or wayland-client types.

src/window_identifier/wayland.rs
```diff
@@ -1,40 +1,79 @@
...
+const ZXDG_EXPORTER_V1: u32 = 1;
+const ZXDG_EXPORTER_V2: u32 = 1;
+
+#[derive(Debug)]
 pub struct WaylandWindowIdentifier {
-    exported: ZxdgExportedV2,
+    exported_v1: Option<ZxdgExportedV1>,
+    exported_v2: Option<ZxdgExportedV2>,
     type_: WindowIdentifierType,
 }
```
i509VCB, line 30: Do we need a `Drop` implementation to destroy the inner `exported` handles?

Do note that destroying exported handles will revoke any imports performed on other clients and the `handle` string will become inert.

A6GibKm: So how it should be handled? Should we expose a method for the user to drop the handle?

src/window_identifier/wayland.rs
```diff
@@ -1,40 +1,79 @@
...
 
+// Supported versions.
+const ZXDG_EXPORTER_V1: u32 = 1;
+const ZXDG_EXPORTER_V2: u32 = 1;
+
+#[derive(Debug)]
 pub struct WaylandWindowIdentifier {
-    exported: ZxdgExportedV2,
+    exported_v1: Option<ZxdgExportedV1>,
+    exported_v2: Option<ZxdgExportedV2>,
```
i509VCB, line 28: We could us an Inner enum where v1 and v2 are variants go make the implementation a bit more idomatic.

src/window_identifier/mod.rs
i509VCB: Actually no we have to have this, RawWindowHandle forces our hand here

src/window_identifier/wayland.rs
i509VCB: gtk3 and gtk4 appear to destroy their exported handles. So yes `Drop` should destroy the exported protocol object.

src/window_identifier/mod.rs
bilelmoussaoui: > Actually no we have to have this, RawWindowHandle forces our hand here

Yes :(

src/window_identifier/wayland.rs
bilelmoussaoui: Yes, I wanted to be consistent with gtk's behaviour. Although, the issue with gtk is it errors out when you try to re-export a handle for an already exported surface instead of returning the same handle. Which make the WindowIdentifier not very nice with gtk, that is why I made it a kind of reference counted from ashpd's side and only unexport when the last usage is dropped.

I am thinking of going directly through the wayland-rs crates in the future and avoid using gtk's provided methods.

bilelmoussaoui:
@i509VCB Do you think this one is ready to be merged for now? we can update later once a new stable release is out.

src/window_identifier/mod.rs
```diff
@@ -27,7 +27,9 @@ use zbus::zvariant::Type;
 ///
 /// ## From a Wayland Surface
 ///
-/// The feature `wayland` must be enabled.
+/// The feature `wayland` must be enabled. The underlying Wayland identifier
+/// will be marked as destroyed on `Drop` and should be keep alive until the
+/// exported surface is no longer in use.
```
bilelmoussaoui, line 32:
```suggestion
/// The `wayland` feature must be enabled. The exported surface handle will be 
/// unexported on `Drop`.
```

bilelmoussaoui:
I will go ahead and land this in order to fix the CI, we can do further changes on another PR if needed.

PR #79, opened 2022-07-19
Pin exact wayland-client versions.

jneem:
I recently tried to use the wayland feature, but it failed because it was pulling in incompatible versions of the various wayland packages. I think the underlying issue is that "0.3.0-beta.3" is not semver compatible with "0.3.0-beta.8", but [cargo thinks it is](git@github.com:bilelmoussaoui/ashpd.git). The fix is to pin the exact versions, and I also had to add an explicit dependency on `wayland-scanner` because otherwise it pulls in `wayland-scanner-0.3.0-beta.8`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,9 +38,10 @@ futures = "0.3"
 tracing = {version = "0.1", optional = true}
 libc = {version = "0.2.94", optional = true}
 raw-window-handle = {version = "0.4", optional = true}
-wayland-client = {version = "0.30.0-beta.3", optional = true}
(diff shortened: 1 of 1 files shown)


bilelmoussaoui:
The wayland feature seems to be broken currently anyway, see #76

jneem:
Ah, indeed! I guess the versions will still need to be pinned after that's sorted out, though.

bilelmoussaoui:
They will, we are waiting for a 0.30 stable release first before merging than and backporting the fixes to 0.2

jneem:
Sounds good! I'll go ahead and close this, then.

PR #80, opened 2022-07-24
Multi camera support

A6GibKm:


src/desktop/camera.rs
```diff
@@ -115,81 +115,135 @@ impl<'a> CameraProxy<'a> {
     }
 }
 
-/// A helper to get the PipeWire Node ID to use with the camera file descriptor
-/// returned by [`CameraProxy::open_pipe_wire_remote`].
+#[cfg(feature = "pipewire")]
+pub struct Stream {
+    pub node_id: u32,
+    pub properties: HashMap<String, String>,
```
bilelmoussaoui, line 121: should be private and have getters like other types

src/desktop/camera.rs
```diff
@@ -115,81 +115,135 @@ impl<'a> CameraProxy<'a> {
     }
 }
 
-/// A helper to get the PipeWire Node ID to use with the camera file descriptor
-/// returned by [`CameraProxy::open_pipe_wire_remote`].
+#[cfg(feature = "pipewire")]
+pub struct Stream {
```
bilelmoussaoui, line 119: Needs docs

ashpd-demo/src/portals/desktop/camera.rs
```diff
@@ -96,10 +99,34 @@ impl CameraPage {
...
-                imp.paintable.set_pipewire_node_id(stream_fd, node_id);
-                imp.revealer.set_reveal_child(true);
-                imp.camera_available.set_text("Yes");
+                let streams = camera::pipewire_streams(stream_fd).await.unwrap();
+                let n_cameras = streams.len();
+                for s in streams.iter() {
+                    let picture = gtk::Picture::new();
+                    let paintable = CameraPaintable::new();
+                    let nick = s.properties.get("node.nick").unwrap();
+                    let name = s.properties.get("node.name").unwrap();
```
bilelmoussaoui, line 108:
```suggestion
                    let nick = s.properties.get("node.nick").unwrap_or_default();
                    let name = s.properties.get("node.name").unwrap_or_default();
```

src/desktop/camera.rs
A6GibKm: Done. Added helpers for `nick`, `description` and `name`. It would be interesting to check whether all camera stream present these properties and unwrap_or_default them instead of returning an Option.

.github/workflows/demo-ci.yml
```diff
@@ -6,6 +6,7 @@ name: Demo CI
 jobs:
   flatpak:
     name: "Flatpak"
+    repository-name: "flathub-beta"
```
bilelmoussaoui, line 9: that is the wrong place to set it :P

src/desktop/camera.rs
```diff
@@ -115,81 +115,161 @@ impl<'a> CameraProxy<'a> {
...
+    let mut map = HashMap::new();
+    for (key, value) in foreign.iter() {
+        map.insert(key.to_string(), value.to_string());
+    }
+    map
+}
+
+#[cfg(feature = "pipewire")]
+/// A PipeWire camera stream returned by [`pipewire_streams`].
+pub struct CameraStream {
```
bilelmoussaoui, line 129: Should be named `Stream` as you are already in the camera module.

src/desktop/camera.rs
```diff
@@ -115,81 +115,161 @@ impl<'a> CameraProxy<'a> {
...
+        self.properties.get("nick").cloned()
+    }
+
+    /// The name of the PipeWire node.
+    pub fn name(&self) -> Option<String> {
+        self.properties.get("name").cloned()
+    }
+
+    /// The description of the PipeWire node.
+    pub fn description(&self) -> Option<String> {
```
bilelmoussaoui, line 157: I am a bit hesitant about providing those getters here tbh, probably worth waiting until we have confirmation about what is common and what we can assume always exists & what camera users would expect to have easily retrievable

A6GibKm: In practice these are the 3 most important properties one cares about, on the 2 devices I tried they were all present.

bilelmoussaoui: sure, the thing is adding new apis in the future is easier than dropping it :P

ashpd-demo/data/resources/ui/camera.ui
```diff
@@ -68,8 +68,20 @@
...
+                <child>
+                  <object class="GtkStackSwitcher" id="picture_stack_switcher">
+                    <property name="stack">picture_stack</property>
+                    <property name="hexpand">True</property>
+                    <property name="halign">center</property>
+                  </object>
+                </child>
+                <child>
+                  <object class="GtkStack" id="picture_stack">
+                  </object>
```
bilelmoussaoui, line 83:
```suggestion
                  <object class="GtkStack" id="picture_stack" />
```

src/desktop/camera.rs
```diff
@@ -115,81 +115,161 @@ impl<'a> CameraProxy<'a> {
...
+    let mut map = HashMap::new();
+    for (key, value) in foreign.iter() {
+        map.insert(key.to_string(), value.to_string());
+    }
+    map
+}
+
+#[cfg(feature = "pipewire")]
+/// A PipeWire camera stream returned by [`pipewire_streams`].
+pub struct Stream {
```
bilelmoussaoui, line 129: needs a `#[derive(Debug)]`

src/desktop/camera.rs
```diff
@@ -115,81 +115,161 @@ impl<'a> CameraProxy<'a> {
...
+}
+
+#[cfg(feature = "pipewire")]
+impl Stream {
+    /// The id of the PipeWire node.
+    pub fn node_id(&self) -> u32 {
+        self.node_id
+    }
+
+    /// Map of node properties.
```
bilelmoussaoui, line 141:
```suggestion
    /// The node properties.
```

bilelmoussaoui:
Looks good enough! thank you

PR #81, opened 2022-07-25
Add multi camera support for the Demo

A6GibKm:


ashpd-demo/src/portals/desktop/camera.rs
```diff
@@ -96,10 +99,35 @@ impl CameraPage {
...
-                imp.revealer.set_reveal_child(true);
-                imp.camera_available.set_text("Yes");
+                let streams = camera::pipewire_streams(stream_fd).await.unwrap();
+                let n_cameras = streams.len();
+                for s in streams.iter() {
+                    let picture = gtk::Picture::new();
+                    let paintable = CameraPaintable::new();
+                    let props = s.properties();
+                    let nick = props.get("nick").map(String::from).unwrap_or_default();
+                    let name = props.get("name").map(String::from).unwrap_or_default();
```
bilelmoussaoui, line 109:
```suggestion
                    let nick = props.get("nick").map(String::from).unwrap_or_default();
                    let name = props.get("name").map(String::from).unwrap_or_default();
```

Seems like the name is often null, at least I got that with the cameras I have on my machine. So I would suggest trying description -> name -> nick.

A6GibKm: The user facing string is nick, so I did `nick` -> `description` -> `name`.

```rust
                    let nick = props
                        .get("nick")
                        .or(props.get("description"))
                        .or(props.get("name"))
                        .map(String::from)
                        .unwrap_or_default();
```

A6GibKm: Note that in that order they go from user-friendly -> device_id.

PR #91, opened 2022-09-15
WIP: add backend support

bilelmoussaoui:


src/backend/account.rs
```diff
@@ -0,0 +1,32 @@
...
+#[dbus_interface(name = "org.freedesktop.impl.portal.Account")]
+impl Account {
+    async fn get_user_information(
+        &self,
+        handle: HandleToken,
+        app_id: &str,
+        window_identifier: WindowIdentifierType,
+        options: UserInfoOptions,
+    ) -> Response<UserInfo> {
+        Response::Err(ResponseError::Cancelled)
```
i509VCB, line 30: Is there a reason Response is like Result instead of modeling the apis around returning a request::Result<UserInfo> type?

bilelmoussaoui: The main reason was the custom deserialize/serialize that are needed for Response;  we can always make that an internal implementation.

bilelmoussaoui:
Closing in favor of #108

PR #93, opened 2022-09-28
Don't directly depend on zvariant

zeenix:
Use it always through zbus.


zeenix:
I'm not sure how can my PR cause the CI breakage: https://github.com/bilelmoussaoui/ashpd/actions/runs/3143398531/jobs/5112284253

bilelmoussaoui:
It is caused by a wayland-rs change, I will fix it later tomorrow

PR #96, opened 2022-11-27
Fix incorrect KeyState mapping in RemoteDesktopProxy

T0b1-iOS:
The KeyState enum for the RemoteDesktopProxy incorrectly maps "Pressed" to 0 and "Released" to 1 when it should be the other way around. (see [here](https://flatpak.github.io/xdg-desktop-portal/#gdbus-method-org-freedesktop-portal-RemoteDesktop.NotifyKeyboardKeycode))
This PR fixes that.


bilelmoussaoui (approved):

bilelmoussaoui:
good catch, thank you!!

PR #97, opened 2022-12-12
Update crates

alatiera:


Cargo.toml
```diff
@@ -37,18 +37,18 @@ pw = {package= "pipewire", version = "0.5", optional = true}
 serde = {version = "1.0", features = ["derive"]}
 serde_repr = "0.1"
 rand = {version = "0.8", default-features = false}
-zbus = {version = "3.1", features = ["url"]}
+zbus = { version = "3.6", features = ["url"] }
```
A6GibKm, line 40: This changes the style of the file.

alatiera: Yea that was intentional for all the lines touched, the spaces is what most of formatting tools default too as well

A6GibKm: I mean, if you want to go this route modify it for all crates on a single commit.

alatiera: meh, let's keep it as is then

bilelmoussaoui:
I actually have these changes locally + some other cleanups, so I would rather wait until I push them first.

bilelmoussaoui:
No longer needed

PR #98, opened 2022-12-12
Paintable GL

alatiera:
- Update crates
- wip: paintable gl update

### DIFF
--- a/ashpd-demo/src/widgets/gst_paintable.rs
+++ b/ashpd-demo/src/widgets/gst_paintable.rs
@@ -96,6 +96,27 @@ impl CameraPaintable {
             .unwrap();
         let paintable = sink.property::<gdk::Paintable>("paintable");
 
+        // create the appropriate sink depending on the environment we are running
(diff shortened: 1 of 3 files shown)


ashpd-demo/src/widgets/gst_paintable.rs
```diff
@@ -96,6 +96,23 @@ impl CameraPaintable {
             .unwrap();
         let paintable = sink.property::<gdk::Paintable>("paintable");
 
+        let (convert, sink) = if paintable
```
A6GibKm, line 99: Is there a reason to define convert here in the tuple? can't the definition be placed before the `let sink = {} ` block?

alatiera: leftover, if pw was giving us a gl/dmabuf buffer we would be using `glcolorconver` instread of `videoconvert`

bilelmoussaoui:
Would you mind rebasing this one please? thank you

PR #99, opened 2022-12-16
Add global shortcuts implementation

PolyMeilex:
Draft because I don't have the setup to test it yet, and it needs some better docs.

~~EDIT: Turns out that it also wrongfully assumes that there is only one `Vec<Shortcut>` type, while in reality there is two, one for defining new shortcuts and one for reading them, so this is also TODO.~~

### DIFF
--- a/src/desktop/global_shortcuts.rs
+++ b/src/desktop/global_shortcuts.rs
@@ -1,0 +1,270 @@
+//! Register global shortcuts
+
+use std::{collections::HashMap, fmt::Debug, time::Duration};
+
+use futures_util::TryFutureExt;
+use serde::{Deserialize, Serialize};
+use zbus::zvariant::{
+    DeserializeDict, ObjectPath, OwnedObjectPath, OwnedValue, SerializeDict, Type,
+};
+
+use super::{HandleToken, Session, DESTINATION, PATH};
+use crate::{
+    helpers::{call_request_method, receive_signal, session_connection},
+    Error, WindowIdentifier,
+};
+
+#[derive(Clone, DeserializeDict, SerializeDict, Type, Debug, Default)]
+#[zvariant(signature = "dict")]
+struct ShortcutInfo {
+    /// User-readable text describing what the shortcut does.
+    description: String,
+    /// User-readable text describing how to trigger the shortcut for the client to render.
+    trigger_description: String,
+}
+
+#[derive(Clone, Serialize, Deserialize, Type, Debug)]
+pub struct Shortcut(String, ShortcutInfo);
+
+impl Shortcut {
+    /// Shortcut id
+    pub fn id(&self) -> &str {
+        &self.0
+    }
+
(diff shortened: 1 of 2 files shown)


bilelmoussaoui:
Thanks a lot for taking the time to implement this!

bilelmoussaoui:
I am looking at doing a new beta in the upcoming days and wonder if you would have the time to update the PR by then? otherwise we can merge it as is and I would push other commits on top of it to finish the implementation/docs

PolyMeilex:
Type mismatch with `Vec<Shortcut>` described in initial description is now fixed. So as far as functionality goes, it should be good now.

src/desktop/global_shortcuts.rs
```diff
@@ -0,0 +1,305 @@
...
+struct NewShortcutInfo {
+    /// User-readable text describing what the shortcut does.
+    description: String,
+    /// The preferred shortcut trigger, defined as described by the "shortcuts" XDG specification. Optional.
+    preferred_trigger: Option<String>,
+}
+
+/// Shortcut descriptor used to bind new shortcuts in [`GlobalShortcuts::bind_shortcuts`]
+#[derive(Clone, Serialize, Type, Debug)]
+pub struct NewShortcut(String, NewShortcutInfo);
```
PolyMeilex, line 28: Not sure if that name is a good choice, but I kinda don't have other ideas, perhaps ShortcutDescriptor or something like that would be better?

src/desktop/global_shortcuts.rs
```diff
@@ -0,0 +1,305 @@
...
+            id.into(),
+            NewShortcutInfo {
+                description: description.into(),
+                preferred_trigger: None,
+            },
+        )
+    }
+
+    /// Sets the preferred shortcut trigger, defined as described by the "shortcuts" XDG specification.
+    pub fn preferred_trigger(mut self, preferred_trigger: impl Into<Option<String>>) -> Self {
```
bilelmoussaoui, line 43:
```suggestion
   #[must_use]
    pub fn preferred_trigger(mut self, preferred_trigger: impl Into<Option<String>>) -> Self {
```

src/desktop/global_shortcuts.rs
bilelmoussaoui: What is actually the difference between NewShortcut & Shortcut? I only see that preferred_trigger is optional and it doesn't make much sense to have the function for creating one have an optional `preferred_trigger` and the other not being optional...

PolyMeilex: When defining new shortcuts, you define, `description` and optionally `preferred_trigger` in XDG defined machine-readable format.

When you read shortcuts already defined previously, you get `description` and `trigger_description`, and both are only human-readable descriptions.

So the common denominator of those two types is `description`.

bilelmoussaoui (approved):

PR #104, opened 2022-12-23
Use safe-io

A6GibKm:
See https://doc.rust-lang.org/src/std/os/fd/owned.rs.html#94

### DIFF
--- a/src/desktop/camera.rs
+++ b/src/desktop/camera.rs
@@ -19,10 +19,10 @@
 
 use std::{
     collections::HashMap,
-    os::unix::prelude::{IntoRawFd, RawFd},
+    os::fd::{FromRawFd, IntoRawFd, OwnedFd},
 };
 
-use zbus::zvariant::{OwnedFd, SerializeDict, Type, Value};
+use zbus::zvariant::{self, SerializeDict, Type, Value};
 
 use super::{HandleToken, DESTINATION, PATH};
 use crate::{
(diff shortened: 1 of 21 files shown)


A6GibKm:
Multiple things need figuring out, for example if pipewire_streams should use OwnedFd instead of BorrowedFd.

A6GibKm:
Also are those unsafe calls Ok? I am sure they will insta crash if the Fd is -1

A6GibKm:
Also requires Rust 1.63

A6GibKm:
Also this one does `fd.close()` on Drop

src/desktop/camera.rs
```diff
@@ -209,13 +212,7 @@ fn pipewire_streams_inner<F: Fn(Stream) + Clone + 'static, G: FnOnce() + Clone +
 /// *Note* The socket referenced by `fd` must not be used while this function is
 /// running.
 #[cfg(feature = "pipewire")]
-pub async fn pipewire_streams(fd: RawFd) -> Result<Vec<Stream>, pw::Error> {
```
bilelmoussaoui, line 212: this means we can drop the dependency on libc right?

A6GibKm: Yes, dropped.

A6GibKm:
Ported the rest of the modules, the only one that requires extra consideration is flatpak/mod.rs since we store the fds of the builder as RawFd, we could store them as OwnedFd.

A6GibKm:
This MR is missing an update to the demo Cargo.toml.

src/flatpak/mod.rs
```diff
@@ -125,28 +120,28 @@ pub struct SpawnOptions {
     /// A list of file descriptor for files inside the sandbox that will be
     /// exposed to the new sandbox, for reading and writing.
     #[zvariant(rename = "sandbox-expose-fd")]
-    sandbox_expose_fd: Option<Vec<Fd>>,
+    sandbox_expose_fd: Vec<Fd<'f>>,
```
bilelmoussaoui, line 123: there is a reason why all those are optional, zbus's SerializeDict doesn't provide a skip_serializing_if functionnality and the portals would explode.

A6GibKm: Ah ok, I thought None was the same as an empty vec

A6GibKm: Using `Option<Vec<Fd<'f>>>` now.

A6GibKm:
There is also PidFd in rust nigthly, I have a small [commit](https://github.com/A6GibKm/ashpd/commit/caebbe206752251236075dcd28904d8e2160094f) for posterity.

bilelmoussaoui:
This PR contains way too many changes, some of them are still interesting but would need a newer one. So closing this one

PR #107, opened 2022-12-30
settings: Fix xdp bug on our side

A6GibKm:
Since future setting keys might be fixed we allow both value(value(x)) and value(x).


bilelmoussaoui (approved):

bilelmoussaoui:
We don't have the CI infra to make sure xdp is actually running so far, so the test is not useful as is.

PR #108, opened 2022-12-31
Add basic backend support

A6GibKm:


src/backend/wallpaper.rs
```diff
@@ -1,36 +1,133 @@
...
+    pub show_preview: Option<bool>,
     #[zvariant(rename = "set-on")]
-    set_on: Option<SetOn>,
+    pub set_on: Option<SetOn>,
+}
+
+pub trait WallpaperImpl {
+    fn set_wallpaper_uri(
+        &self,
+        handle: OwnedObjectPath,
```
bilelmoussaoui, line 24: Ideally we should pass in a Request created from this path btw

Cargo.toml
```diff
@@ -51,6 +51,7 @@ async-std = {version = "1.12", optional = true}
 tokio = {version = "1.21", features = ["fs", "io-util"], optional = true, default-features = false}
 once_cell = "1.14"
 url = {version = "2.3", features = ["serde"]}
+async-trait = "0.1.60"
```
bilelmoussaoui, line 54: should be enabled only when the backend option is enabled

src/backend/mod.rs
```diff
@@ -3,3 +3,5 @@ mod account;
 mod request;
 mod session;
 mod wallpaper;
+
+pub use crate::desktop::HandleToken;
```
bilelmoussaoui: Should be exposed only when the backend option is enabled

src/backend/wallpaper.rs
```diff
@@ -0,0 +1,143 @@
...
+        app_id: &str,
+        window_identifier: WindowIdentifierType,
+        uri: url::Url,
+        options: WallpaperOptions,
+    ) -> u32;
```
bilelmoussaoui, line 29:
```suggestion
    ) -> BasicResponse;
```
instead, make that type public along with `Response` if backend is enabled then, add two constructors to `BasicResponse` for `cancelled`  / `other`  /  `success` or `default`

src/backend/file_chooser.rs
```diff
@@ -0,0 +1,289 @@
...
+#[async_trait]
+pub trait FileChooserImpl {
+    async fn open_file(
+        &self,
+        handle: OwnedObjectPath,
+        app_id: &str,
+        window_identifier: WindowIdentifierType,
+        title: &str,
+        options: OpenFileOptions,
+    ) -> (u32, OpenFileResults);
```
bilelmoussaoui, line 86:
```suggestion
    ) -> Response<OpenFileResults>;
```
and elsewhere

src/backend/wallpaper.rs
```diff
@@ -0,0 +1,143 @@
...
+    mpsc::{Receiver, Sender},
+    oneshot,
+};
+use futures_util::{FutureExt, StreamExt};
+use std::cell::RefCell;
+use zbus::{dbus_interface, names::WellKnownName};
+
+use crate::{backend::IMPL_PATH, desktop::wallpaper::SetOn, WindowIdentifierType};
+
+#[derive(DeserializeDict, Type, Debug, Default)]
```
bilelmoussaoui, line 12:
```suggestion
#[derive(DeserializeDict, Type, Debug)]
```
The Default is not needed here

src/backend/request.rs
```diff
@@ -1,10 +1,94 @@
...
+    pub fn new(sender: Sender<Action>) -> Self {
+        Self { sender }
+    }
+}
 
 #[dbus_interface(name = "org.freedesktop.impl.portal.Request")]
-impl Request {
-    async fn close(&self) -> zbus::fdo::Result<()> {
+impl RequestInterface {
+    #[dbus_interface(name = "SetRequestURI")]
```
bilelmoussaoui, line 84: is this needed?

src/backend/request.rs
```diff
@@ -1,10 +1,96 @@
...
+#[dbus_interface(name = "org.freedesktop.impl.portal.Request")]
+impl RequestInterface {
+    async fn close(&mut self) -> BasicResponse {
+        let (sender, receiver) = futures_channel::oneshot::channel();
+        let _ = self.sender.try_send(Action::Close(sender));
+        let mut stream = receiver.into_stream();
+        let next = stream.next().await;
+        next.unwrap().unwrap();
+
+        BasicResponse::default()
```
bilelmoussaoui, line 94: i am sure this method doesn't expect any response?

src/desktop/wallpaper.rs
```diff
@@ -52,6 +52,7 @@ use crate::{
 
 #[derive(Serialize, Deserialize, Debug, Clone, Copy, PartialEq, Eq, Hash, Type)]
 #[zvariant(signature = "s")]
+#[serde(rename_all = "lowercase")]
```
bilelmoussaoui, line 55: Is this needed only here ? might be worth checking other enums as well while you are at it.

src/backend/file_chooser.rs
```diff
@@ -0,0 +1,293 @@
...
+    pub current_name: Option<String>,
+    pub current_folder: Option<Vec<u8>>,
+    pub current_file: Option<Vec<u8>>,
+}
+
+#[derive(DeserializeDict, Type, Debug, Default)]
+#[zvariant(signature = "dict")]
+pub struct SaveFilesOptions {
+    // TODO Its in the xdp docs, but is it correct? See
+    // https://github.com/flatpak/xdg-desktop-portal/issues/938
```
bilelmoussaoui, line 52: This is probably a docs issue, the handle_token doesn't make much sense here and should be dropped.

src/backend/file_chooser.rs
```diff
@@ -0,0 +1,293 @@
...
+    pub modal: Option<bool>,
+    pub choices: Option<Vec<Choice>>,
+    pub current_folder: Option<Vec<u8>>,
+    pub files: Option<Vec<Vec<u8>>>,
+}
+
+#[derive(DeserializeDict, SerializeDict, Type, Debug, Default)]
+#[zvariant(signature = "dict")]
+pub struct OpenFileResults {
+    pub uris: Option<Vec<url::Url>>,
```
bilelmoussaoui, line 64: All those pub fields should get getters instead.

src/desktop/wallpaper.rs
A6GibKm: I checked all the pub enums and I think its only needed there.

src/backend/file_chooser.rs
A6GibKm: Ill leave the comment and comment 
```rust
    // pub handle_token: Option<String>,
```

src/desktop/request.rs
```diff
@@ -31,6 +31,31 @@ where
...
+    }
+
+    pub fn other() -> Self {
+        Self::Err(ResponseError::Other)
+    }
+}
+
+#[cfg(feature = "backend")]
+impl Response<BasicResponse> {
+    pub fn empty() -> Self {
```
bilelmoussaoui, line 54: I wonder if we  can use Default here instead?

src/desktop/file_chooser.rs
```diff
@@ -125,6 +125,31 @@ impl FileFilter {
     }
 }
 
+#[cfg(feature = "backend")]
+impl FileFilter {
+    pub fn label(&self) -> &str {
+        &self.0
+    }
+
+    pub fn mimetype_filters(&self) -> Vec<&str> {
```
bilelmoussaoui, line 134:
```suggestion
    pub fn mimetypes(&self) -> Vec<&str> {
```

src/desktop/file_chooser.rs
```diff
@@ -125,6 +125,31 @@ impl FileFilter {
...
+        self.1
+            .iter()
+            .filter_map(|(type_, string)| {
+                (type_ == &FilterType::MimeType).then_some(string.as_str())
+            })
+            .collect()
+    }
+
+    pub fn pattern_filters(&self) -> Vec<&str> {
```
bilelmoussaoui, line 143:
```suggestion
    pub fn patterns(&self) -> Vec<&str> {
```

src/desktop/file_chooser.rs
bilelmoussaoui: also those new getters could use a bit of docs

src/backend/settings.rs
```diff
@@ -0,0 +1,101 @@
...
+unsafe impl<T: Send + SettingsImpl> Send for Settings<T> {}
+unsafe impl<T: Sync + SettingsImpl> Sync for Settings<T> {}
+
+impl<T: SettingsImpl> Settings<T> {
+    pub async fn new(imp: T, backend: &Backend) -> zbus::Result<Self> {
+        let (sender, receiver) = futures_channel::mpsc::channel(10);
+        let iface = SettingsInterface::new(sender);
+        let object_server = backend.cnx().object_server();
+
+        object_server.at(IMPL_PATH, iface).await?;
```
bilelmoussaoui, line 37:
```suggestion
        let object_server = backend.cnx().object_server();

        object_server.at(IMPL_PATH, iface).await?;
       
```
this seems to be repeated everywhere, can you add a helper method for it to the Backend type that would do the necessary calls for you?

src/backend/mod.rs
```diff
@@ -0,0 +1,86 @@
...
+
+        proxy
+            .request_name(
+                name.clone(),
+                zbus::fdo::RequestNameFlags::ReplaceExisting.into(),
+            )
+            .await?;
+
+        Ok(Backend {
+            cnx: Some(cnx),
```
bilelmoussaoui, line 54: Why are those nullable?

bilelmoussaoui: nvm

Cargo.toml
```diff
@@ -50,6 +51,8 @@ async-std = {version = "1.12", optional = true}
 tokio = {version = "1.21", features = ["fs", "io-util"], optional = true, default-features = false}
 once_cell = "1.14"
 url = {version = "2.3", features = ["serde"]}
+async-trait = {version = "0.1.60", optional = true}
+futures = { version = "0.3.25", features = ["executor", "thread-pool"], optional = true}
```
bilelmoussaoui, line 55: do we actually need the whole futures crate here?

src/backend/mod.rs
```diff
@@ -15,3 +19,68 @@ pub use settings::{Settings, SettingsImpl};
...
+    cnx: Option<zbus::Connection>,
+    name: Option<WellKnownName<'static>>,
+}
+
+impl Backend {
+    pub async fn new<N: TryInto<WellKnownName<'static>>>(name: N) -> Result<Self, crate::Error>
+    where
+        zbus::Error: From<<N as TryInto<WellKnownName<'static>>>::Error>,
+    {
+        Self::new_inner(name).await.map_err(From::from)
```
bilelmoussaoui, line 35: What is the purpose of new_inner? nothing seems to use it ...

src/backend/mod.rs
```diff
@@ -15,3 +19,68 @@ pub use settings::{Settings, SettingsImpl};
...
+    fn cnx(&self) -> &zbus::Connection {
+        self.cnx.as_ref().unwrap()
+    }
+}
+
+impl Drop for Backend {
+    fn drop(&mut self) {
+        if let (Some(cnx), Some(name)) = (self.cnx.take(), self.name.take()) {
+            let executor = ThreadPool::new().unwrap();
+            if let Err(_err) = executor.spawn(async move {
```
bilelmoussaoui, line 75: We already have an async-std/tokio features, wouldn't be more correct to use those here instead of depending on the futures crate?

src/desktop/request.rs
A6GibKm: I dont thin the semantics fit

Cargo.toml
```diff
@@ -12,6 +12,7 @@ version = "0.4.0-alpha.1"
 
 [features]
 default = ["async-std"]
+backend = ["async-trait", "futures"]
```
bilelmoussaoui, line 15: The new feature has to be documented in the readme file

src/backend/mod.rs
A6GibKm: The purpose is to have something that returns a zbus::Result to use `?`, the tryinto for N is a bit wonky.

Cargo.toml
A6GibKm: only for the spawn, but if we already depend always on either tokio or async-std then we can remove it.

src/desktop/file_chooser.rs
A6GibKm: We need to change the api here https://docs.rs/ashpd/latest/ashpd/desktop/file_chooser/struct.FileFilter.html first otherwise its a bit confusing.

src/backend/file_chooser.rs
```diff
@@ -46,8 +86,8 @@ pub struct SaveFileOptions {
     pub current_filter: Option<FileFilter>,
     pub choices: Option<Vec<Choice>>,
     pub current_name: Option<String>,
-    pub current_folder: Option<Vec<u8>>,
-    pub current_file: Option<Vec<u8>>,
+    pub current_folder: Option<FileName>,
```
bilelmoussaoui, line 89: You should also replace functions making use of  path_from_null_terminated

src/backend/file_chooser.rs
```diff
@@ -22,6 +26,42 @@ use crate::{
...
+    }
+}
+
+impl Type for FileName {
+    fn signature() -> zbus::zvariant::Signature<'static> {
+        <&[u8]>::signature()
+    }
+}
+
+impl<'de> Deserialize<'de> for FileName {
```
bilelmoussaoui, line 52: Please also implement Serialize so that we can  drop stuff like https://github.com/bilelmoussaoui/ashpd/blob/master/src/desktop/file_chooser.rs#L443-L447.

A6GibKm: In the pr it has serialize, fortunately we don't need a custom serialize.

ashpd-backend-demo/src/external_window.rs
```diff
@@ -0,0 +1,51 @@
+use ashpd::WindowIdentifierType;
+use gtk::prelude::*;
+use gtk::{gdk, glib};
+
+use crate::external_wayland_window::ExternalWaylandWindow;
+use crate::external_x11_window::ExternalX11Window;
+
+pub(crate) enum ExternalWindow {
```
bilelmoussaoui, line 8: I believe this enum is useless. All we need is to `impl From<WindowIdentifierType>  for gtk::Window`, no need to split into separate structs, could be just separate functions that do all the calls for you.

bilelmoussaoui: Once that is done, please move it to `window_identifier` module inside `ashpd` itself and behind `gtk3_wayland`/`gtk3_x11` / `gtk4_wayland` / `gtk4_x11` features.

bilelmoussaoui: That is a part that we could merge separately from this PR in order to simplify the review process of the backend.

src/app_id.rs
```diff
@@ -7,15 +7,36 @@ use zbus::zvariant::Type;
 #[derive(Debug, Serialize, Deserialize, Type, PartialEq, Eq, Hash)]
 pub struct AppID(String);
 
+unsafe impl Send for AppID {}
```
bilelmoussaoui, line 10: Could use a rebase

src/desktop/account.rs
```diff
@@ -34,29 +34,38 @@ struct UserInformationOptions {
...
 
-#[derive(Debug, DeserializeDict, Type)]
+#[derive(Debug, Default, SerializeDict, DeserializeDict, Type)]
 /// The response of a [`UserInformationRequest`] request.
 #[zvariant(signature = "dict")]
 pub struct UserInformation {
-    id: String,
-    name: String,
-    image: url::Url,
+    id: Option<String>,
```
bilelmoussaoui, line 41: Why are those now nullable? I don't believe they are, are they?

A6GibKm: We needed to have Default there, I don't remember why. Url does not have default so it needs an option, maybe we can only wrap that one in option. In any case for me it makes more sense to pass None than `""`.

ashpd-backend-demo/src/external_window.rs
A6GibKm: I think its more involved, you need to create a window, parent the surface and keep a display.

src/desktop/account.rs
bilelmoussaoui: When would we have a None here? I don't understand why Default is even needed here.

src/backend/file_chooser.rs
```diff
@@ -0,0 +1,335 @@
...
+    zvariant::{DeserializeDict, OwnedObjectPath, SerializeDict, Type},
+    AppID, WindowIdentifierType,
+};
+
+// FIXME Manually implement Deserialize to catch non null-terminated arrays.
+// FIXME Use OsString directly once
+// https://gitlab.freedesktop.org/dbus/zbus/-/merge_requests/638 is merged.
+/// A file name represented as a null-terminated byte array.
+///
+/// Implements `AsRef<Path>`.
```
bilelmoussaoui, line 34: Could be dropped now

src/desktop/account.rs
A6GibKm: This was reverted.

ashpd-backend-demo/src/screenshot.rs
```diff
@@ -0,0 +1,127 @@
...
+};
+use async_trait::async_trait;
+
+mod shell_screenshot {
+    use zbus::dbus_proxy;
+
+    use super::Color;
+
+    #[dbus_proxy(
+        interface = "org.gnome.Shell.Screenshot",
```
jadahl, line 20: Please don't use any private GNOME APIs in components that are not version locked with the rest of GNOME. These APIs are not stable and will only work for components in the same release.

ashpd-backend-demo/src/main.rs
```diff
@@ -0,0 +1,124 @@
...
+    gettextrs::bindtextdomain("ashpd-backend-demo", "/usr/share/locale")
+        .expect("Unable to bind the text domain");
+    gettextrs::textdomain("ashpd-backend-demo").expect("Unable to switch to the text domain");
+
+    glib::set_prgname(Some("ashpd-backend-demo"));
+
+    // Avoid pointless and confusing recursion
+    glib::unsetenv("GTK_USE_PORTAL");
+    glib::setenv("ADW_DISABLE_PORTAL", "1", true).unwrap();
+    glib::setenv("GSK_RENDERER", "cairo", true).unwrap();
```
jadahl, line 40: This has really really bad performance on gtk4 in GNOME 44 so should be avoided. With that said, it's not possible to have GdkDisplays with different backends while still using hardware acceleration. See https://gitlab.gnome.org/GNOME/xdg-desktop-portal-gnome/-/merge_requests/68.

ashpd-backend-demo/src/screenshot.rs
bilelmoussaoui: This is barely a demo to try out the API, we are not planning to make use of that anywhere, so I think that is fine ?

jadahl: That demo will only work if you run the shell with the "unsafe" mode activated, until it changes the next time, unless you pretend to be a component you're not, but it still raises some warning flags to use these APIs outside the platform. Does a demo really need to use "real" implementations to demonstrate the API, instead of just returning some predefined image?

ashpd-backend-demo/src/main.rs
A6GibKm: I am confused by this, we need x11 displays to have modal windows on windows using xwalyand. Does this means that xdp-gnome does not do modals on them anymore? By the way note that initializing gtk with the GL backaend takes extremely long (~0.5s-1s on my machines), not a big problem if it is only done once but still.

jadahl: xdp-gnome 44 uses Wayland for modals on top of both Wayland and X11 windows if it's a Wayland session, and X11 otherwise. It uses a [custom protocol](https://gitlab.gnome.org/GNOME/xdg-desktop-portal-gnome/-/blob/main/data/mutter-x11-interop.xml) just for that.

orowith2os:
Is this functional enough to try out with a custom portal, or should I instead look to making my own helpers on top of zbus?

ashpd-backend-demo/src/screenshot.rs
bilelmoussaoui: Agree with what you said here. The scope of this PR went too big that it would take forever to ever land. I think we should just focus on providing building blocks of writing portals. Keep the demo as basic as possible.

bilelmoussaoui:
We have diverged quite a lot in main and simplified the whole process, it should be much simpler to add new portals backend support. Regarding the demo, I think we should keep it as simple as possible, so bringing gtk is a bit too much. 

Thanks for starting this initiative though, closing.

PR #109, opened 2023-01-04
window_identifier: Do not destroy exported surface

A6GibKm:
The other side of the bus should take ownership of the exported toplevel and destroy it, otherwise things might go badly. This is something that should be called only once the associated GtkWindow (for example) is destroyed but we don't have a nice api for this.

One can check that destroying the exported toplevel too quick causes the file chooser to not be modal, but due to sheer luck it works with the current implementation of the file chooser, other portals might not be as lucky.

We also use Option<String> to transport the handle, this ensures we don't have an empty string comming from Default::default().


A6GibKm:
For reference https://gitlab.gnome.org/GNOME/gtk/-/issues/5495.

src/window_identifier/wayland.rs
```diff
@@ -24,25 +24,9 @@ const ZXDG_EXPORTER_V2: u32 = 1;
...
     type_: WindowIdentifierType,
 }
 
-#[derive(Debug)]
-enum Exported {
-    V1(ZxdgExportedV1),
-    V2(ZxdgExportedV2),
-}
-
-impl Exported {
```
bilelmoussaoui, line 37: I feel like we should provide some way for the user to destroy those whenever they want

A6GibKm:
Closing as I am not sure anymore this solves anything.

PR #114, opened 2023-01-11
Use a custom struct for nul-terminated strings

A6GibKm:


A6GibKm:
+40 -96, big win. Only question remaining is if we should keep those `.expect("no null terminated stuff")` around.

bilelmoussaoui:
You forgot to push a file

A6GibKm:
> You forgot to push a file

Done.

src/desktop/file_chooser.rs
```diff
@@ -522,22 +516,18 @@ impl SaveFileRequest {
     /// Sets the current folder.
     #[must_use]
     pub fn current_folder<P: AsRef<Path>>(mut self, current_folder: impl Into<Option<P>>) -> Self {
```
bilelmoussaoui, line 518: Could those use `TryInto<FileName>` instead?

A6GibKm: I think not, we cannot implement `impl<S: AsRef<Path>> TryFrom<S> for FileName` since it conflicts with others.

src/lib.rs
```diff
@@ -23,7 +23,7 @@ mod window_identifier;
 pub use self::window_identifier::WindowIdentifier;
 mod app_id;
 pub use self::app_id::AppID;
-mod file_name;
+mod file_path;
```
bilelmoussaoui, line 26: This also needs a `pub use self::file_path::FilePath;` and also please squash the commits that adds the type and the one that renames it

src/file_name.rs
```diff
@@ -0,0 +1,42 @@
...
+use std::{
+    ffi::{CString, OsStr},
+    os::unix::ffi::OsStrExt,
+    path::Path,
+};
+use zbus::zvariant::Type;
+
+/// A file name represented as a null-terminated byte array.
+///
+/// Implements `AsRef<Path>`.
```
bilelmoussaoui: No need for this part I believe. it is already in the docs. 

Also please move this file to the src/ directory as it is common with various "sub-directories" and would also be used by the backend.

src/documents/mod.rs
```diff
@@ -32,22 +32,15 @@
...
+use std::{collections::HashMap, fmt, os::unix::prelude::AsRawFd, path::Path, str::FromStr};
 
 use enumflags2::{bitflags, BitFlags};
 use serde::{Deserialize, Serialize};
 use serde_repr::{Deserialize_repr, Serialize_repr};
 use zbus::zvariant::{Fd, OwnedValue, Type};
 
 pub use crate::app_id::DocumentID;
-use crate::{helpers::path_from_null_terminated, proxy::Proxy, AppID, Error};
+use crate::{file_path::FilePath, proxy::Proxy, AppID, Error};
```
bilelmoussaoui, line 43:
```suggestion
use crate::{FilePath, proxy::Proxy, AppID, Error};
```
and everywhere else

A6GibKm: Thanks, done.

A6GibKm:
By the way, I used nul-terminated everywhere in code following rust

bilelmoussaoui (approved):

A6GibKm:
Rebased

bilelmoussaoui (approved):

PR #115, opened 2023-01-12
Changes to AppID and DocumentID

A6GibKm:


src/app_id.rs
```diff
@@ -34,3 +60,89 @@ impl From<String> for DocumentID {
...
+    c.is_ascii_alphabetic() || matches!(c, '-' | '_')
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_is_valid_app_id() {
+        assert!(is_valid_app_id("a.b"));
```
bilelmoussaoui, line 134: Do you mind also adding the missing test cases from https://github.com/bilelmoussaoui/flatpak-vscode/blob/master/src/test/suite/extension.test.ts#L143-L241?

A6GibKm: Done, but we fail one test that had `-` outside the last segment so I removed that one (it did not test anything else interesting).

PR #120, opened 2023-02-15
Add getters for FileFilter

A6GibKm:


src/desktop/file_chooser.rs
```diff
@@ -125,6 +125,30 @@ impl FileFilter {
...
+impl FileFilter {
+    pub fn label(&self) -> &str {
+        &self.0
+    }
+
+    pub fn mimetype_filters(&self) -> Vec<&str> {
+        self.1
+            .iter()
+            .filter_map(|(type_, string)| {
+                (type_ == &FilterType::MimeType).then_some(string.as_str())
```
bilelmoussaoui, line 137:
```suggestion
                (type_ == &FilterType::MimeType).then_some(string.as_str())
```

In this case I prefer if you could add two methods to `FilterType` named `is_mimetype` and `is_pattern` and make use of them here.

A6GibKm: Check the current code.

bilelmoussaoui (approved):

PR #125, opened 2023-03-22
cargo: don't depend on zbus' async-io feature when we use tokio

Cogitri:


Cogitri:
Couldn't test this just yet though

bilelmoussaoui:
Let us assume if it builds on CI, it works. Will do some more testing later

Issue #129, opened 2023-05-12
`Documents::add_full` errors with `InvalidAppID` if `app_id` is empty string


ids1024:
This function is supposed to take an app ID **or** an empty string, but `TryFrom<String>` calls `is_valid_app_id`, which returns `false` for empty strings.

This should also apply to `add_named_full` and `list`.

Would it be better to skip using a `TryFrom` generic here, and accept `Option<AppId>`? That would probably be more idiomatic, and let the caller deal with (fallible) conversion.

bilelmoussaoui:
> Would it be better to skip using a TryFrom generic here, and accept Option<AppId>?

Mind submitting a patch?

PR #132, opened 2023-05-15
documents: Take `AppId` or `OptionAppId` instead of `TryInto` generic

ids1024:
Some of these functions accept an empty string, but calling with that was erroring since those are not valid as app ids. So accepting an `Option` would be idiomatic in Rust.

It could take `Option<T>` where `T` has a `TryFrom` bound, but that's an awkward API to deal with, since `None` can't be passed without a type. It also seems unidiomatic to take a generic argument when the conversion is failable; it would be cleaner to leave failable conversion to the caller.

Fixes https://github.com/bilelmoussaoui/ashpd/issues/129.

### DIFF
--- a/src/documents/mod.rs
+++ b/src/documents/mod.rs
@@ -8,7 +8,7 @@
 //!
 //!     println!("{:#?}", proxy.mount_point().await?);
 //!
-//!     for (doc_id, host_path) in proxy.list("org.mozilla.firefox").await? {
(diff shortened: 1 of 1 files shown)


bilelmoussaoui (approved):

bilelmoussaoui:
lgtm, thanks!! 

I suppose you need a release with those changes?

PR #138, opened 2023-06-21
Duplicate all fds we have as input

A6GibKm:
This way we don't run into the case where the fd is closed on the app side before we are done with it. By using OwnedFd, we make sure the dupped fd is closed when it goes out of scope, after the async operation is done.

Note that the only api missing is the attach in email.rs.

### DIFF
--- a/src/helpers.rs
+++ b/src/helpers.rs
@@ -1,7 +1,11 @@
+use std::os::unix::prelude::{AsRawFd, BorrowedFd, OwnedFd};
+
 #[cfg(feature = "async-std")]
 use async_std::{fs::File, prelude::*};
 #[cfg(feature = "tokio")]
 use tokio::{fs::File, io::AsyncReadExt};
+
+use crate::Error;
 
 pub(crate) async fn is_flatpak() -> bool {
     #[cfg(feature = "async-std")]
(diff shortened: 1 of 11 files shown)


bilelmoussaoui:
Is this part https://github.com/bilelmoussaoui/ashpd/blob/master/src/desktop/camera.rs#L195 still needed? If not it should be removed along the libc dependency

A6GibKm:
good catch, ill update it tomorrow

src/helpers.rs
```diff
@@ -51,6 +55,12 @@ fn cgroup_v2_is_snap(cgroups: &str) -> bool {
         .any(|x| x.unwrap_or(false))
 }
 
+pub(crate) fn dup_to_owned_fd(fd: &impl AsRawFd) -> Result<OwnedFd, Error> {
```
bilelmoussaoui, line 58: Maybe make this return a zvariant::FD?

A6GibKm: The point of using OwnedFd is that it will automatically be closed on drop and will correctly do the duplication with `F_DUPFD_CLOEXEC`, FD doesn't and won't do this (their OwnedFd does close on Drop tho)

bilelmoussaoui: I am just not sure why such helper exists and yet all the times we use it we wrap it in a zvariant type. So I wonder if we could do things slightly differently

A6GibKm: well, ideally zbus learns how to serialize `OwnedFd` and `BorrowedFd<'_>`, no need to turn them into `zbus::Fd` or `zbus::OwnedFd`

A6GibKm:
Marked as draft until I can test the demo

bilelmoussaoui:
We now take OwnedFd / BorrowedFd depends on the use cases, that streamlined the API a little bit. Closing as the purpose of this PR is no longer needed and we would just need to make sure we are not mis-using BorrowedFd/OwnedFd in certain cases

PR #150, opened 2023-08-07
demo: enable the ashpd pipewire feature

whot:
Fixes
```
error[E0425]: cannot find function `pipewire_streams` in module `camera`
  --> src/portals/desktop/camera.rs:96:39
   |
96 |                 let streams = camera::pipewire_streams(stream_fd).await.unwrap();
   |                                       ^^^^^^^^^^^^^^^^ not found in `camera`

```
Missing the `Cargo.lock` changes though, see #149

PR #151, opened 2023-08-07
Allow running the demo from the build dir

whot:
This sits on top of #150 so it builds on my machine but can be split out if need be.

Unless there's a trick I'm missing, we can't easily run the demo from the meson build dir because it looks in `/usr/local/share/` for the gresources file. This could (possibly?) be worked around by setting the datadir in meson but another hack is to build the same binary with a `.devel` suffix, check for that suffix in `argv0` and load the gresources from the builddir instead. Not overly pretty but for this use-case it's probably good enough.

### DIFF
--- a/ashpd-demo/src/main.rs
+++ b/ashpd-demo/src/main.rs
@@ -6,7 +6,7 @@ mod widgets;
 mod window;
 
 use application::Application;
-use config::{GETTEXT_PACKAGE, LOCALEDIR, RESOURCES_FILE};
+use config::{GETTEXT_PACKAGE, LOCALEDIR, RESOURCES_FILE, BUILDDIR_RESOURCES_FILE};
 use gettextrs::*;
 use gtk::{gio, glib};
 
@@ -24,7 +24,12 @@ fn main() -> glib::ExitCode {
 
     gst4gtk::plugin_register_static().expect("Failed to register gstgtk4 plugin");
 
-    let res = gio::Resource::load(RESOURCES_FILE).expect("Could not load gresource file");
+    let argv0 = std::env::args().next();
+    let res = if argv0.is_some() && argv0.unwrap().ends_with(".devel") {
+        gio::Resource::load(BUILDDIR_RESOURCES_FILE)
+    } else {
+        gio::Resource::load(RESOURCES_FILE)
+    }.expect("Could not load gresource file");
     gio::resources_register(&res);
 
     let mut args = std::env::args();
--- a/ashpd-demo/src/config.rs.in
+++ b/ashpd-demo/src/config.rs.in
@@ -5,3 +5,4 @@ pub const PKGDATADIR: &str = @PKGDATADIR@;
 pub const PROFILE: &str = @PROFILE@;
 pub const RESOURCES_FILE: &str = concat!(@PKGDATADIR@, "/resources.gresource");
 pub const VERSION: &str = @VERSION@;
+pub const BUILDDIR_RESOURCES_FILE: &str = concat!(@BUILDDATADIR@, "/resources.gresource");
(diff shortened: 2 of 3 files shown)


bilelmoussaoui:
Could use a rebase

whot:
Rebased but switched back to draft. I tried running this on a new machine and luckily caught the error: the current branch still requires that the schemas are installed. Simply adding `XDG_DATA_DIRS` isn't enough, we need to also compile the schemas in the builddir but that's where I ran into a bit of a rabbit hole. I'm not sure when I'll find time to complete this :(

ashpd-demo/src/main.rs
```diff
@@ -24,7 +24,12 @@ fn main() -> glib::ExitCode {
 
     gst4gtk::plugin_register_static().expect("Failed to register gstgtk4 plugin");
 
-    let res = gio::Resource::load(RESOURCES_FILE).expect("Could not load gresource file");
+    let argv0 = std::env::args().next();
+    let res = if argv0.is_some() && argv0.unwrap().ends_with(".devel") {
+        gio::Resource::load(BUILDDIR_RESOURCES_FILE)
+    } else {
+        gio::Resource::load(RESOURCES_FILE)
+    }.expect("Could not load gresource file");
```
bilelmoussaoui, line 32: I am not particularly happy with this approach of checking the binary name. I think a viable path forward is to have a separate meson build option that would influence from where we load the gresource file making the changes to config.rs/main.rs not needed at all. The other option is to embed the gresource into the final binary but nobody took the time to figure out a nice way to do it that integrates well with the hackish meson/cargo thing we are doing

whot:
> For the schemas, you can use GSETTINGS_SCHEMA_DIR

So from what I've gathered so far
- `gio::Resource::load()` apparently requires a full path to the file, so we need to hack around that anyway
- `gio::Settings::new()` requires *compiled* schemas but meson's [gnome.compile_schemas()](https://mesonbuild.com/Gnome-module.html#gnomecompile_schemas) can only compile in the source directory. We generate the `gschema.xml` files so they are in the build directory. That seems like a strangely missing part of the meson gnome module and needs to be fixed there. Manually compiling in the builddir and setting `GSETTINGS_SCHEMA_DIR` works.

If I hack around the two above, I get a segfault somewhere deep inside GTK4's `gdk_wayland_toplevel_focus()` which I'm assuming is not ashpd's fault :)

But I agree, the hackish way right now is not nice and since it doesn't work anyway, let's close this PR - I'm not sure when I'll have time to look at this again in the near future.

eli-schwartz:
> * `gio::Settings::new()` requires _compiled_ schemas but meson's [gnome.compile_schemas()](https://mesonbuild.com/Gnome-module.html#gnomecompile_schemas) can only compile in the source directory. We generate the `gschema.xml` files so they are in the build directory. That seems like a strangely missing part of the meson gnome module and needs to be fixed there. Manually compiling in the builddir and setting `GSETTINGS_SCHEMA_DIR` works.

It feels a bit awkward, mostly because glib-compile-schemas only compiles a single directory, and we hacked it up with --target to place the compiled version in a different tree hierarchy. I don't know that it's feasible to merge the two together, so we'd either need to pick one, or generate another directory as an internal implementation detail and append to the GSETTINGS_SCHEMA_DIR directory.

Of course this is all nothing users can't do manually, the main value of the module is to guarantee it is correct...

whot:
Thanks for the explanation! From what I can tell from a quick browse `glib-compile-schemas` does take a srcdir argument (`argv[1]`), and I suspect during autotools builds srcdir == builddir in most cases anyway (also given the need for `--target` to be added), so my random guess is that passing an optional srcdir to the module to be passed through would *probably* work? but that's just that, a random guess :)

Issue #148, opened 2023-08-07
Receive color scheme changed does not work


d2weber:
The function
https://github.com/bilelmoussaoui/ashpd/blob/471bae1fb9376ae92f572e0c91c428149ec89190/src/desktop/settings.rs#L161
does not seem to work. The following program never returns.

```toml
# Cargo.toml
[dependencies]
ashpd = "0.4.0"
```

```rust
// src/main.rs
use ashpd::{desktop::settings, zbus::block_on};

fn main() {
    block_on(async {
        let settings = settings::Settings::new().await.unwrap();
        let s = settings.receive_color_scheme_changed().await.unwrap();
        dbg!(s);
    });
}
```

I'm using Ubuntu 22.04.2 LTS (GNOME 42.9)

The loop in `receive_color_scheme_changed` waits for a changed setting, checks if it matches it's filters and if it doeesn't, checks again for a changed setting. I suspect that there are settings that are not matching the filter criteria and while the checks are being made, the _desired_ setting changed message arrives. But they are gone by the time that the loop will check for changed settings again.

I also don't quite understand the concept of hiding away the `Stream` interface of the signals, see https://github.com/bilelmoussaoui/ashpd/blob/471bae1fb9376ae92f572e0c91c428149ec89190/src/proxy.rs#L157 Why isn't the stream returned directly? The user can then call `next().await` and discard the stream to achieve the current behaviour of `receive_setting_changed` and `receive_color_scheme_changed`. I also tried to find any uses of this interface in the demo folder and the rest of github but I couldn't find any.

I believe it would be useful for the user to get the stream, I have an implementation [here](https://github.com/d2weber/tintanum/blob/main/src/lib.rs#L86-L94) and I would submit a PR to integrate it in ashpd if it is desired.

bilelmoussaoui:
> I believe it would be useful for the user to get the stream, I have an implementation [here](https://github.com/d2weber/tintanum/blob/main/src/lib.rs#L86-L94) and I would submit a PR to integrate it in ashpd if it is desired.

Patches would be welcome, only if all the other signals are replaced with returning an actual stream

PR #152, opened 2023-08-09
Expose streams for signals

d2weber:
#148

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,6 +25,7 @@ wayland = ["wayland-client", "wayland-protocols", "wayland-backend"]
 async-std = {version = "1.12", optional = true}
 enumflags2 = "0.7"
 futures-channel = "0.3"
+futures-lite = "1.13.0"
 futures-util = "0.3"
 gdk4wayland = {package = "gdk4-wayland", version = "0.8", git="https://github.com/gtk-rs/gtk4-rs.git", optional = true}
 gdk4x11 = {package = "gdk4-x11", version = "0.8", git="https://github.com/gtk-rs/gtk4-rs.git", optional = true}
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,7 +25,6 @@ wayland = ["wayland-client", "wayland-protocols", "wayland-backend"]
 async-std = {version = "1.12", optional = true}
 enumflags2 = "0.7"
 futures-channel = "0.3"
-futures-lite = "1.13.0"
 futures-util = "0.3"
 gdk4wayland = {package = "gdk4-wayland", version = "0.8", git="https://github.com/gtk-rs/gtk4-rs.git", optional = true}
 gdk4x11 = {package = "gdk4-x11", version = "0.8", git="https://github.com/gtk-rs/gtk4-rs.git", optional = true}
--- a/src/desktop/settings.rs
+++ b/src/desktop/settings.rs
@@ -187,34 +187,6 @@ impl<'a> Settings<'a> {
             .await
     }
 
-    /// Listen to changes of the system's preferred color scheme
-    ///
-    /// The stream will also initially yield the current color scheme preference and filter out
-    /// duplicates.
-    pub async fn receive_color_scheme_changed_with_init(
-        &self,
-    ) -> Result<impl Stream<Item = ColorScheme>, Error> {
-        let mut current = self.color_scheme().await?;
-        Ok(once(ready(current)).chain(
-            self.0
-                .signal_with_args::<Setting>(
-                    "SettingChanged",
-                    &[(0, APPEARANCE_NAMESPACE), (1, COLOR_SCHEME_KEY)],
-                )
-                .await?
-                .filter_map(move |p| {
-                    ready(ColorScheme::try_from(p).ok().and_then(|p| {
-                        if p == current {
-                            None
-                        } else {
-                            current = p;
-                            Some(p)
-                        }
-                    }))
-                }),
-        ))
-    }
-
     /// Signal emitted when a setting changes.
     ///
     /// # Specifications
--- a/src/desktop/settings.rs
+++ b/src/desktop/settings.rs
@@ -27,10 +27,10 @@
 //! }
 //! ```
 
-use std::future::ready;
+
 use std::{collections::HashMap, convert::TryFrom, fmt::Debug};
 
-use futures_util::{stream::once, Stream, StreamExt};
+use futures_util::{Stream};
 use serde::{Deserialize, Serialize};
 use zbus::zvariant::{OwnedValue, Type, Value};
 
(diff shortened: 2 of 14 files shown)


d2weber:
I tested that the color scheme works. But I don't know how to make sure that the other signals work. Unfortunately I didn't manage to build the demo because gio seems to be outdated: 
```
  error: could not find system library 'gio-2.0' required by the 'gio-sys' crate

  --- stderr
  Requested 'gio-2.0 >= 2.76' but version of GIO is 2.72.4
```
I believe these version requirements are outdated:
https://github.com/bilelmoussaoui/ashpd/blob/a1fee8bd65ea4f5bf4b7639b1ccc8531b06205d4/ashpd-demo/meson.build#L13-L15

Cargo.toml
```diff
@@ -25,6 +25,7 @@ wayland = ["wayland-client", "wayland-protocols", "wayland-backend"]
 async-std = {version = "1.12", optional = true}
 enumflags2 = "0.7"
 futures-channel = "0.3"
+futures-lite = "1.13.0"
```
bilelmoussaoui, line 28: Nothing in the commit message documents the reasoning behind this switch

src/desktop/settings.rs
```diff
@@ -181,14 +181,27 @@ impl<'a> Settings<'a> {
     }
 
     /// Listen to changes of the system's preferred color scheme
-    pub async fn receive_color_scheme_changed(
+    ///
+    /// The stream will also initially yield the current color scheme preference and filter out
+    /// Duplicates.
+    pub async fn receive_color_scheme_changed_with_init(
```
bilelmoussaoui, line 187: `duplicates` instead also not really a fan of this function tbh. The user can easily call `self.color_scheme().await?;` and listen to changes afterwards.

src/desktop/settings.rs
```diff
@@ -73,6 +74,37 @@ pub enum ColorScheme {
...
+impl TryFrom<Setting> for ColorScheme {
+    type Error = Error;
+
+    fn try_from(value: Setting) -> Result<Self, Self::Error> {
+        Self::try_from(value.2)
+    }
+}
+
+const NAMESPACE: &str = "org.freedesktop.appearance";
+const KEY: &str = "color-scheme";
```
bilelmoussaoui, line 107: Rename this to `COLOR_SCHEME_KEY`

src/desktop/settings.rs
```diff
@@ -73,6 +74,37 @@ pub enum ColorScheme {
...
+
+impl TryFrom<Setting> for ColorScheme {
+    type Error = Error;
+
+    fn try_from(value: Setting) -> Result<Self, Self::Error> {
+        Self::try_from(value.2)
+    }
+}
+
+const NAMESPACE: &str = "org.freedesktop.appearance";
```
bilelmoussaoui, line 106: and this to `APPEARANCE_NAMESPACE`

bilelmoussaoui:
> I tested that the color scheme works. But I don't know how to make sure that the other signals work. Unfortunately I didn't manage to build the demo because gio seems to be outdated:
> 
> ```
>   error: could not find system library 'gio-2.0' required by the 'gio-sys' crate
> 
>   --- stderr
>   Requested 'gio-2.0 >= 2.76' but version of GIO is 2.72.4
> ```
> 
> I believe these version requirements are outdated:
> 
> https://github.com/bilelmoussaoui/ashpd/blob/a1fee8bd65ea4f5bf4b7639b1ccc8531b06205d4/ashpd-demo/meson.build#L13-L15


I can update the demo later, but at least the tests should be fixed

Thank you for the patches!

Cargo.toml
d2weber: The reason is that futures-lites `filter_map` takes a simple FnMut that maps to an Option but the one from futures-utils ([docs](https://docs.rs/futures-util/latest/futures_util/stream/trait.StreamExt.html#method.filter_map)) takes an FnMut that maps to a future, resolving to the output type. That makes the implementation harder, and even when I made it compile, when I actually tried to use the resulting streams next method, the compiler complained that the steam wasn't pin. (I believe it would require to the stream)

I can include the reasoning in a commit message.

d2weber: Note that throughout the codebase two different versions of `SteamExt` will be used. I'm open to suggestions to clean that up. But not sure which would be the best way to go.

src/desktop/settings.rs
d2weber: The alternative to initializing `current` would be using an Option type. I liked the Idea to get the initial value in the stream, because I used the signal to call a function that makes sure everything is setup with the current preference and I don't need to store the preference in a variable. In my case I just had to wait for a changed color-scheme and execute my function and that's it. Would be interesting how other usecases look like. Or if you prefer I do it with the Option (because I believe duplicates are not nice).

Cargo.toml
d2weber: No it's also used by `signal()` in `proxy.rs` for the type conversion.

Not sure what you mean by better semantic. Do you mean the `impl Stream` part?

d2weber:
I fixed the things you mentioned. In the demo there will be at least these issues:
https://github.com/bilelmoussaoui/ashpd/blob/a1fee8bd65ea4f5bf4b7639b1ccc8531b06205d4/ashpd-demo/src/portals/desktop/location.rs#L167
https://github.com/bilelmoussaoui/ashpd/blob/a1fee8bd65ea4f5bf4b7639b1ccc8531b06205d4/ashpd-demo/src/portals/desktop/inhibit.rs#L112
https://github.com/bilelmoussaoui/ashpd/blob/a1fee8bd65ea4f5bf4b7639b1ccc8531b06205d4/ashpd-demo/src/portals/desktop/notification.rs#L131

Cargo.toml
bilelmoussaoui: I don't think that you need futures_lite for filter_map ,see https://github.com/bilelmoussaoui/oo7/pull/40 where i used it without depending on that new dependency.

d2weber: You are totally right, I don't need futures_lite. I just realized, that futures_utils' version of `filter_map` is more general, because you can easily convert a value to a future using `ready`. I feel a little stupid that I understood it only now. I was trying to make an async block with all the logic and moving a string holding the interface name into the created future. That complicated matters a lot.
Sidenote: futures_lite is already in the dependecy tree trough zbus itself and async_std. (Of course I understand that you want to minimize direct dependencies though)

bilelmoussaoui: Right, people could complain to zbus/asynd-std if they want. Thanks for the fixes!

src/desktop/settings.rs
bilelmoussaoui: This is the only remaining "problem". I would suggest we drop this part from the patches and open a separate PR for it so we can get the signal stream fixes in, that would also simplify the review/discussion

d2weber: What would 'drop' mean, removing the function?

bilelmoussaoui: yes and open a separate patch with only that function

bilelmoussaoui:
Clippy is complaining

bilelmoussaoui:
Thank you!

PR #153, opened 2023-08-10
Add receive_color_scheme_changed_with_init

d2weber:
Follow up of #152 
Only works if rebased

### DIFF
--- a/src/desktop/settings.rs
+++ b/src/desktop/settings.rs
@@ -30,7 +30,7 @@
 use std::future::ready;
 use std::{collections::HashMap, convert::TryFrom, fmt::Debug};
 
-use futures_util::{stream::once, Stream, StreamExt};
+use futures_util::{Stream, StreamExt};
 use serde::{Deserialize, Serialize};
 use zbus::zvariant::{OwnedValue, Type, Value};
 
@@ -188,31 +188,17 @@ impl<'a> Settings<'a> {
     }
 
     /// Listen to changes of the system's preferred color scheme
-    ///
-    /// The stream will also initially yield the current color scheme preference and filter out
-    /// duplicates.
-    pub async fn receive_color_scheme_changed_with_init(
+    pub async fn receive_color_scheme_changed(
         &self,
     ) -> Result<impl Stream<Item = ColorScheme>, Error> {
-        let mut current = self.color_scheme().await?;
-        Ok(once(ready(current)).chain(
-            self.0
-                .signal_with_args::<Setting>(
-                    "SettingChanged",
-                    &[(0, APPEARANCE_NAMESPACE), (1, COLOR_SCHEME_KEY)],
-                )
-                .await?
-                .filter_map(move |p| {
-                    ready(ColorScheme::try_from(p).ok().and_then(|p| {
-                        if p == current {
-                            None
-                        } else {
-                            current = p;
-                            Some(p)
-                        }
-                    }))
-                }),
-        ))
+        Ok(self
+            .0
+            .signal_with_args::<Setting>(
+                "SettingChanged",
+                &[(0, APPEARANCE_NAMESPACE), (1, COLOR_SCHEME_KEY)],
+            )
+            .await?
+            .filter_map(|x| ready(ColorScheme::try_from(x).ok())))
     }
 
     /// Signal emitted when a setting changes.
--- a/src/desktop/settings.rs
+++ b/src/desktop/settings.rs
@@ -187,6 +187,34 @@ impl<'a> Settings<'a> {
             .await
     }
 
+    /// Listen to changes of the system's preferred color scheme
+    ///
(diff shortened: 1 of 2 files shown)


d2weber:
TODO: Revert 'Remove unused' once #152 is merged

bilelmoussaoui:
Needs a rebase

src/desktop/settings.rs
```diff
@@ -186,6 +187,34 @@ impl<'a> Settings<'a> {
             .await
     }
 
+    /// Listen to changes of the system's preferred color scheme
+    ///
+    /// The stream will also initially yield the current color scheme preference and filter out
+    /// duplicates.
+    pub async fn receive_color_scheme_changed_with_init(
```
bilelmoussaoui, line 194: So two questions here:

- What is really the use case of yielding the current color scheme?

- What does filtering out the duplicates mean? how can the signal be emitted if the setting didn't change, it should not. If it happens, it is a backend issue and should not be worked around from a frontend library.

src/proxy.rs
```diff
@@ -157,6 +157,31 @@ impl<'a> Proxy<'a> {
             .map_err(From::from)
     }
 
+    pub(crate) async fn signal_with_args<I>(
+        &self,
+        name: &'static str,
+        args: &[(u8, &str)],
```
bilelmoussaoui, line 163: can't the args also be `[(u8, &'static str)]`?

src/desktop/settings.rs
d2weber: > What is really the use case of yielding the current color scheme?

The way I would use it in my cli application is waiting for a changed color-scheme and perform an action when it changes. Not sure how it would be used in a bigger (gui) application.


> What does filtering out the duplicates mean? how can the signal be emitted if the setting didn't change, it should not. If it happens, it is a backend issue and should not be worked around from a frontend library.

That's a good question. I just experienced that on my machine the signal always gets emitted twice. A google search wasn't helpful. But probably you're right and we shouldn't handle this issue here.

src/proxy.rs
d2weber: I don't understand what you mean, I don't think it's possible because array wouldn't have a known size at compile time. The dbus function `receive_signal_with_args` also takes a slice `&[(u8, &str)]`. I don't think it makes sense to make this function generic over different arraysizes.

src/desktop/settings.rs
bilelmoussaoui: Which xdg-desktop-portal you have installed? Also could you use something like Bustle, record the dbus traffic and share the pcap file here so we can investigate which part of the stack is not behaving as expected?

Regarding the cli app you are working on, I think the best would be to have a closure and call it on the initial state as well as when the state changes. That would be much cleaner then the current API suggestion.

src/proxy.rs
bilelmoussaoui: Never mind, I should have enough coffee when reviewing stuff next time

src/desktop/settings.rs
d2weber: > Which xdg-desktop-portal you have installed? Also could you use something like Bustle, record the dbus traffic and share the pcap file here so we can investigate which part of the stack is not behaving as expected?

Thank you, that you want to investigate this case. I'm using Ubuntu 22.04 but I'm not sure which backend is used. Both gnome.portal  gtk.portal support the "Settings" portal and are `UseIn=gnome`, `$XDG_SESSION_DESKTOP` is `ubuntu`. So I guess it would be (because it comes first in the alphabet) gnome.portal? Here the [pcap]([2023-08-11 15:27:03.zip](https://github.com/bilelmoussaoui/ashpd/files/12321956/2023-08-11.15.27.03.zip)

> Regarding the cli app you are working on, I think the best would be to have a closure and call it on the initial state as well as when the state changes. That would be much cleaner then the current API suggestion.

Yes that's definitely doable.

d2weber: Looking at the pcap it could be that both gtk and gnome are active, which might be the root cause. But I don't know how I could have misconfigured my system. Maybe because I installed additional desktop environments? It's also not crucial to fix this on my machine..

bilelmoussaoui: > Looking at the pcap it could be that both gtk and gnome are active, which might be the root cause. But I don't know how I could have misconfigured my system. Maybe because I installed additional desktop environments? It's also not crucial to fix this on my machine..

You can have multiple backends, each of them providing only a subset of the portals. It is normal to have both gtk & gnome ones.

bilelmoussaoui:
Thanks a lot for your patches! I will do a new 0.6 release soonish with your changes

d2weber:
Thank you for your help!

bilelmoussaoui:
There is now a 0.6 release with your fixes

d2weber:
Great! Thank you 👍

PR #162, opened 2023-08-29
settings: Add support for accent-color

A6GibKm:


A6GibKm:
The biggest change on Color was the addition of the `From<OwnedValue>` otherwise
```
error[E0277]: the trait bound `color::Color: From<OwnedValue>` is not satisfied
   --> src/desktop/settings.rs:178:14
    |
178 |         self.read::<Color>(APPEARANCE_NAMESPACE, ACCENT_COLOR_SCHEME_KEY)
    |              ^^^^ the trait `From<OwnedValue>` is not implemented for `color::Color`
    |
    = note: required for `OwnedValue` to implement `Into<color::Color>`
    = note: required for `color::Color` to implement `TryFrom<OwnedValue>`

For more information about this error, try `rustc --explain E0277`.
warning: `ashpd` (lib) generated 2 warnings
```

src/color.rs
```diff
@@ -0,0 +1,78 @@
+use crate::zvariant::{self, DeserializeDict, Type};
+use crate::{desktop::screenshot::ColorRequest, Error};
+
+#[derive(DeserializeDict, Clone, Copy, PartialEq, Type)]
+/// The response of a [`ColorRequest`] request.
```
bilelmoussaoui, line 5: Update the comment

src/color.rs
```diff
@@ -0,0 +1,78 @@
...
+    /// Blue.
+    pub fn blue(&self) -> f64 {
+        self.color[2]
+    }
+
+    /// Creates a new builder-pattern struct instance to construct
+    /// [`Color`].
+    ///
+    /// This method returns an instance of [`ColorRequest`].
+    pub fn request() -> ColorRequest {
```
bilelmoussaoui, line 33: This no longer make much sense now, maybe we should figure out how to make it agnostic between settings and color picker?

src/desktop/settings.rs
```diff
@@ -172,6 +173,12 @@ impl<'a> Settings<'a> {
         }
     }
 
+    /// Retrieves the system's preferred accent color
+    pub async fn accent_color(&self) -> Result<Color, Error> {
```
bilelmoussaoui, line 177: Please add receive_accent_color_changed as well

src/color.rs
A6GibKm: We don't need it to be agnostic. Maybe we can just add a `impl Color` with this single method in the color picker's file

bilelmoussaoui: I would keep it in the same file and have a new method that returns the accent color?

src/color.rs
A6GibKm: Done.

src/desktop/screenshot.rs
```diff
@@ -172,6 +172,16 @@ impl ColorRequest {
     }
 }
 
+impl Color {
+    /// Creates a new builder-pattern struct instance to construct
+    /// [`Color`].
+    ///
+    /// This method returns an instance of [`ColorRequest`].
+    pub fn pick() -> ColorRequest {
```
bilelmoussaoui, line 180: Not enough, the tests would fail :p

A6GibKm: Updated everywhere, except the demo

src/color.rs
```diff
@@ -0,0 +1,72 @@
+use crate::{
```
bilelmoussaoui, line 1: This file should be inside desktop as it is specific to there. Similar to say HandleToken and so on

bilelmoussaoui:
Thank you

PR #163, opened 2023-08-29
Add versioned methods

A6GibKm:
We throw an error on our side when an interface is called from our side with the wrong version.

TODO
  - [ ] One should add documentation to the methods requiring a newer version.
  - [x] Cache the version property


src/proxy.rs
```diff
@@ -122,6 +122,10 @@ impl<'a> Proxy<'a> {
         self.request(handle_token, method_name, body).await
     }
 
+    pub async fn version(&self) -> Result<u32, Error> {
+        self.property::<u32>("version").await
+    }
+
```
bilelmoussaoui, line 128: This should be called once for the said proxy and stored in it struct so we don't have to fetch the value every time

A6GibKm: Done, should I transform the struct to
```rust
Proxy  {
    version: u32,
    inner: ...
}
```
too? or just `Proxy(x,y)` is enough?

bilelmoussaoui: I would use what you suggested here to make it clearer.

bilelmoussaoui: Would you mind documenting the version requirements? the next step, not necessarily to be done as part of this PR is to implement Deref for the inner proxy so we can make version available to all the portal objects.

bilelmoussaoui: I have an idea how to do that, so don't bother with it :)

src/proxy.rs
```diff
@@ -58,13 +58,15 @@ impl<'a> Proxy<'a> {
         P::Error: Into<zbus::Error>,
     {
         let connection = Self::connection().await?;
-        let proxy = zbus::ProxyBuilder::new_bare(&connection)
+        let proxy: zbus::Proxy = zbus::ProxyBuilder::new_bare(&connection)
```
bilelmoussaoui, line 61:
```suggestion
        let proxy = zbus::ProxyBuilder::new_bare(&connection)
```
not needed

src/proxy.rs
```diff
@@ -122,6 +124,10 @@ impl<'a> Proxy<'a> {
         self.request(handle_token, method_name, body).await
     }
 
+    pub async fn version(&self) -> u32 {
```
bilelmoussaoui, line 127:
```suggestion
    pub fn version(&self) -> u32 {
```
no longer async & the various awaits are not needed

src/proxy.rs
A6GibKm: Ok, I used the struct tho

src/proxy.rs
A6GibKm: I get 
```
error[E0282]: type annotations needed
  --> src/proxy.rs:64:13
   |
64 |         let inner = zbus::ProxyBuilder::new_bare(&connection)
   |             ^^^^^
...
70 |         let version = inner.get_property::<u32>("version").await.unwrap_or(1);
   |                       ----- type must be known at this point
   |
help: consider giving `inner` an explicit type
   |
64 |         let inner: /* Type */ = zbus::ProxyBuilder::new_bare(&connection)
   |                  ++++++++++++

```
otherwise

src/proxy.rs
A6GibKm: Right

PR #165, opened 2023-09-02
Add missing options

A6GibKm:


src/desktop/email.rs
```diff
@@ -167,6 +168,13 @@ impl EmailRequest {
         self
     }
 
+    /// Sets the activation token.
+    #[must_use]
+    pub fn activation_token<'a>(mut self, activation_token: impl Into<Option<&'a str>>) -> Self {
```
bilelmoussaoui, line 173: Do we expose that everywhere else as well?

A6GibKm:
The print portal commit was marked as wip since this won't be exposed to the user, probably the `prepare_print` fn should be set to use a builder pattern otherwise it has too many optional arguments.

src/desktop/email.rs
A6GibKm: Nope.

bilelmoussaoui: Then, we should for consistency or mark it as as private with a todo item saying we should handle it once we support passing activation

bilelmoussaoui:
> The print portal commit was marked as wip since this won't be exposed to the user, probably the `prepare_print` fn should be set to use a builder pattern otherwise it has too many optional arguments.

I didn't spend time to make the print portal using builder pattern because it is too much.

A6GibKm:
> > The print portal commit was marked as wip since this won't be exposed to the user, probably the `prepare_print` fn should be set to use a builder pattern otherwise it has too many optional arguments.
> 
> I didn't spend time to make the print portal using builder pattern because it is too much.

So just add a new argument?

bilelmoussaoui:
> > > The print portal commit was marked as wip since this won't be exposed to the user, probably the `prepare_print` fn should be set to use a builder pattern otherwise it has too many optional arguments.
> > 
> > 
> > I didn't spend time to make the print portal using builder pattern because it is too much.
> 
> So just add a new argument?

For now yes

bilelmoussaoui:
Could you drop the WIP from the last commit ? Otherwise lgtm

A6GibKm:
https://github.com/flatpak/xdg-desktop-portal/pull/1094 hasn't been merged yet so...

PR #170, opened 2023-10-29
Correct a comment and a typo

carlosmn:
The first change is that `Spawn` is not about running things outside the sandbox but inside. There is a different interface for running processes outside of the sandbox.

The second one is an all-to-common typo of writing `existed` instead of `exited`. This does break API but it also fixes it so I don't know what you'd want to do about accepting that. Maybe the typoed version should be deprecated instead?

### DIFF
--- a/src/flatpak/mod.rs
+++ b/src/flatpak/mod.rs
@@ -1,6 +1,6 @@
 //! # Examples
 //!
-//! Spawn a process outside of the sandbox, only works in a Flatpak.
+//! Spawn a process inside of the sandbox, only works in a Flatpak.
 //!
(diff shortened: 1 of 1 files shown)


bilelmoussaoui:
I don't plan to do a new .x release anytime soon, so it is fine to break API. As we would bump to zbus 4.0 soon-ish anyways

bilelmoussaoui:
If you need the fixed version, you can always cherry-pick the changes and mark the wrongly named one as deprecated in the 0.6 branch

carlosmn:
That's all fine then.

PR #172, opened 2023-11-03
Enable gtk4 conditionally on x11/wayland instead of the other way around

### LINKED ISSUE #171: Change dependency ordering in Cargo.toml across gtk4/gtk4x11/gtk4wayland
Currently, ashpd's cargo.toml selects gtk4wayland and gtk4x11 both if gtk4 is enabled.

This is an issue when trying to achieve x11/wayland only builds. Explicitly selecting said features, disables `WindowIdentifier::from_native` which is hidden behind the gtk4 feature

This is specially problematic when trying to build downstream applications such as gnome-authenticator for a single DM

Changing this dependency order, allows for wayland/x11/both builds, while retaining the rest of the functionaliy as is.

ambasta:
Fixes #171

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,9 +14,9 @@ rust-version = "1.65"
 [features]
 async-std = ["zbus/async-io", "dep:async-std"]
 default = ["async-std"]
-gtk4 = ["gtk4_x11", "gtk4_wayland"]
-gtk4_wayland = ["gdk4wayland", "dep:gtk4"]
-gtk4_x11 = ["gdk4x11", "dep:gtk4"]
+gtk4 = []
+gtk4_wayland = ["gtk4", "gdk4wayland", "dep:gtk4"]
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -14,9 +14,9 @@ rust-version = "1.65"
 [features]
 async-std = ["zbus/async-io", "dep:async-std"]
 default = ["async-std"]
-gtk4 = ["gtk4_x11", "gtk4_wayland"]
-gtk4_wayland = ["gdk4wayland", "dep:gtk4"]
-gtk4_x11 = ["gdk4x11", "dep:gtk4"]
+gtk4 = []
```
bjorn, line 17: Judging by the build errors, maybe this should be `gtk4 = ["dep:gtk4"]`, and then this dep can be removed from the other two?

bilelmoussaoui:
This was properly fixed by a different PR

PR #176, opened 2023-12-07
flatpak: add the Development interface to launch programs on the host

carlosmn:
This is a lot like spawn but you sometimes need to run programs from the host system. This is e.g. used by GNOME Builder to run flatpak itself to build your stuff.

I don't know how much sense it makes to get the demo app to use this as it would require adding host filesystem access for it to really show something.

### DIFF
--- a/src/proxy.rs
+++ b/src/proxy.rs
@@ -21,6 +21,9 @@ pub(crate) const DOCUMENTS_PATH: &str = "/org/freedesktop/portal/documents";
 pub(crate) const FLATPAK_DESTINATION: &str = "org.freedesktop.portal.Flatpak";
 pub(crate) const FLATPAK_PATH: &str = "/org/freedesktop/portal/Flatpak";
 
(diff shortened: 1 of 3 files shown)


src/proxy.rs
```diff
@@ -21,6 +21,9 @@ pub(crate) const DOCUMENTS_PATH: &str = "/org/freedesktop/portal/documents";
 pub(crate) const FLATPAK_DESTINATION: &str = "org.freedesktop.portal.Flatpak";
 pub(crate) const FLATPAK_PATH: &str = "/org/freedesktop/portal/Flatpak";
 
+pub(crate) const FLATPAK_DEVELOPMENT_DESTINATION: &str = "org.freedesktop.Flatpak";
```
bilelmoussaoui, line 24: I would name them FLATPAK_HOST instead

bilelmoussaoui (approved):
Looks good to me, we might need to split what we have inside flatpak/mod.rs into flatpak/sandbox.rs or something like that in the future but that is outside of the scope of this PR. Thanks!

bilelmoussaoui:
I will land this as is, the remaining changes are minor.

PR #177, opened 2023-12-18
input_capture: Fix parsing of `CreateSessionResponse`, `Zones`

ids1024:


bilelmoussaoui:
Thanks! Sadly I couldn't test it at all, do you have a use of it somewhere?

ids1024:
I've just updated https://github.com/ids1024/reis/blob/master/examples/receive.rs to make use of it. That should work on a current version of Gnome.

Just the part using the portal would be enough to test for issues like this, even without actually using the ei fd.

PR #186, opened 2024-01-03
Fix possible OnceCell panic in Proxy::connection()

### LINKED ISSUE #185: Simultaneous creation of proxies results in a "Can't reset a OnceCell" panic
The following minimal example can reliably reproduce the issue:
```rust
use ashpd::desktop::{input_capture::InputCapture, remote_desktop::RemoteDesktop};

#[tokio::main(flavor="current_thread")]
async fn main() {
    tokio::join!(ic(), rdp());
}

async fn ic() {
    let _ = InputCapture::new().await.unwrap();
}

async fn rdp() {
    let _ = RemoteDesktop::new().await.unwrap();
}
```
Repo with code in question: [feschber/ashpd-mre](https://github.com/feschber/ashpd-mre)

The problem appears to be that [Proxy::connection()](https://github.com/bilelmoussaoui/ashpd/blob/1da9ed3d85d7cfc4b76f602bbcb597f958e971bd/src/proxy.rs#L40) does not check if the OnceCell is already initialized after awaiting the creation of the new connection which results in a "Can't resest a OnceCell" panic.

feschber:
When multiple tasks are trying to create proxies simultaneously, another task may initialize the zbus-Connection-OnceCell while the connection is initialized.

This ensures that only the first task to establish a connection initializes the OnceCell

Fixes #185

### DIFF
--- a/src/proxy.rs
+++ b/src/proxy.rs
@@ -38,8 +38,8 @@ impl<'a> Proxy<'a> {
             Ok(cnx.clone())
         } else {
             let cnx = zbus::Connection::session().await?;
-            SESSION.set(cnx.clone()).expect("Can't reset a OnceCell");
-            Ok(cnx)
+            // during `await` another task may have initialized the cell
+            Ok(SESSION.get_or_init(|| { cnx }).clone())
         }
     }
 
--- a/src/proxy.rs
+++ b/src/proxy.rs
@@ -39,7 +39,7 @@ impl<'a> Proxy<'a> {
         } else {
             let cnx = zbus::Connection::session().await?;
             // during `await` another task may have initialized the cell
-            Ok(SESSION.get_or_init(|| { cnx }).clone())
+            Ok(SESSION.get_or_init(|| cnx).clone())
         }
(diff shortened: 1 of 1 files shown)


A6GibKm:
I am not particularly happy with this solution, ideally we turn the entire fn into a get_or_init, but you need to get the session in an async way which is incompatible with Lazy and OnceCell... I wonder if using std::sync::Once helps here.

feschber:
I understand what you mean but I don't think there is a trivial solution to this problem that would ensure the connection is initialized only once.

To do it more cleanly, one would need to atomically check if the cell is already being initialized and then wait for the initializing task to finish somehow. And then there is the potential problem of that task being cancelled as well.
Basically an async get_or_init I guess.

There is [https://github.com/danieldg/async-once-cell](https://github.com/danieldg/async-once-cell) which does all of that but it seems a bit complex and I'm not sure if it is worth pulling in this dependency just for this.

bilelmoussaoui:
Rustfmt is complaining

PR #193, opened 2024-02-18
Take impl AsFd instead of BorrowedFd

A6GibKm:
This is more general, in particular the following will still compile.

```rust
let fd = something_implementing_as_fd;
let fd: &std::os::fd::BorrowedFd<'_> = &fd.as_fd();
ashpd::any_method_using_a_fd(fd).await?;
```

### DIFF
--- a/src/desktop/game_mode.rs
+++ b/src/desktop/game_mode.rs
@@ -15,7 +15,7 @@
 //! }
 //! ```
 
-use std::{fmt::Debug, os::fd::BorrowedFd};
+use std::{fmt::Debug, os::fd::AsFd};
 
 use serde_repr::Deserialize_repr;
 use zbus::zvariant::{Fd, Type};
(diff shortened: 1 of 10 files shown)


A6GibKm:
Something annoying is that with this you cannot just use `HashMap::new`  when one of generics is `impl AsFd`, see the docs change.

bilelmoussaoui:
Needs a rebase

A6GibKm:
Done.

src/desktop/game_mode.rs
```diff
@@ -113,13 +113,13 @@ impl<'a> GameMode<'a> {
     #[doc(alias = "QueryStatusByPIDFd")]
     pub async fn query_status_by_pidfd(
         &self,
-        target: &BorrowedFd<'_>,
-        requester: &BorrowedFd<'_>,
+        target: impl AsFd,
+        requester: impl AsFd,
```
bilelmoussaoui, line 117: Why not take &impl AsFd instead? If you will just end up borrowing

A6GibKm: Changed it to `&impl AsFd`.

bilelmoussaoui (approved):

PR #196, opened 2024-03-04
0.7 accent

wash2:
I believe the accent color is defined [here](https://github.com/flatpak/xdg-desktop-portal/blob/main/data/org.freedesktop.impl.portal.Settings.xml) . Without the patch, it panics trying to parse the value, but maybe I've misinterpreted the type defined in the xml.

### DIFF
--- a/src/desktop/color.rs
+++ b/src/desktop/color.rs
@@ -13,6 +13,10 @@ pub struct Color {
 }
 
 impl Color {
+    pub(crate) fn new(r: f64, g: f64, b: f64) -> Self {
+        Self { color: [r, g, b] }
+    }
+
     /// Red.
     pub fn red(&self) -> f64 {
         self.color[0]
--- a/src/desktop/settings.rs
+++ b/src/desktop/settings.rs
@@ -205,8 +205,9 @@ impl<'a> Settings<'a> {
 
     /// Retrieves the system's preferred accent color
     pub async fn accent_color(&self) -> Result<Color, Error> {
-        self.read::<Color>(APPEARANCE_NAMESPACE, ACCENT_COLOR_SCHEME_KEY)
+        self.read::<(f64, f64, f64)>(APPEARANCE_NAMESPACE, ACCENT_COLOR_SCHEME_KEY)
             .await
+            .map(|(r, g, b)| Color::new(r, g, b))
     }
 
     /// Retrieves the system's preferred color scheme
@@ -232,13 +233,14 @@ impl<'a> Settings<'a> {
     }
 
     /// Listen to changes of the system's accent color
-    pub async fn receive_accent_color_changed(
-        &self,
-    ) -> Result<impl Stream<Item = ColorScheme>, Error> {
+    pub async fn receive_accent_color_changed(&self) -> Result<impl Stream<Item = Color>, Error> {
         Ok(self
-            .receive_setting_changed_with_args(APPEARANCE_NAMESPACE, ACCENT_COLOR_SCHEME_KEY)
+            .receive_setting_changed_with_args::<(f64, f64, f64)>(
+                APPEARANCE_NAMESPACE,
+                ACCENT_COLOR_SCHEME_KEY,
+            )
             .await?
-            .filter_map(|t| ready(t.ok())))
+            .filter_map(|t| ready(t.ok().map(|(r, g, b)| Color::new(r, g, b)))))
     }
 
     /// Listen to changes of the system's contrast level


src/desktop/settings.rs
```diff
@@ -232,9 +232,7 @@ impl<'a> Settings<'a> {
     }
 
     /// Listen to changes of the system's accent color
-    pub async fn receive_accent_color_changed(
-        &self,
-    ) -> Result<impl Stream<Item = ColorScheme>, Error> {
+    pub async fn receive_accent_color_changed(&self) -> Result<impl Stream<Item = Color>, Error> {
```
A6GibKm, line 235: whats the justification of this change?

A6GibKm:
Could you please explain what do you mean by

> Without the patch, it panics trying to parse the value, but maybe I've misinterpreted the type defined in the xml.

Did the api crash when testing against a portal impl that supported this key, or that seems to be the case?

src/desktop/settings.rs
wash2: The `ColorScheme` type is not the correct type. I explained in the commit that changes it that it just seems like a typo. The accent-color value is a color. `ColorScheme` is meant for the color-scheme value.
```
pub enum ColorScheme {
    NoPreference,
    PreferDark,
    PreferLight,
}
```

wash2:
> Could you please explain what do you mean by
> 
> > Without the patch, it panics trying to parse the value, but maybe I've misinterpreted the type defined in the xml.
> 
> Did the api crash when testing against a portal impl that supported this key, or that seems to be the case?

Yes it crashes against to cosmic implementation of the portal. I've tested specifically with this https://github.com/pop-os/xdg-desktop-portal-cosmic/pull/28. I could test with more implementations to be sure though.

https://dbus.freedesktop.org/doc/dbus-specification.html#container-types and https://github.com/flatpak/xdg-desktop-portal/blob/04fdd3579abef01248997608e2abd919be94c1b1/data/org.freedesktop.impl.portal.Settings.xml#L43 lead me to believe that the type is a struct, not a dict.

src/desktop/color.rs
```diff
@@ -1,31 +1,31 @@
...
 
-#[derive(DeserializeDict, Clone, Copy, PartialEq, Type)]
-/// A color as a RGB tuple.
+#[derive(
+    Default, DeserializeDict, Clone, Copy, PartialEq, Type, zvariant::Value, zvariant::OwnedValue,
+)]
+/// A color in the sRGB color space.
 ///
 /// **Note** the values are normalized in the [0.0, 1.0] range.
-#[zvariant(signature = "dict")]
```
bilelmoussaoui, line 10: This is wrong.

wash2: Sorry, I don't understand. Do you mean the line: `#[zvariant(signature = "dict")]` is wrong? I can revert my changes though if it is not a struct. It's just that I interpreted (ddd) as a struct type with three f64.

bilelmoussaoui: The current code is wrong because it is shared between screenshot's pick color & the accent color. The first one expects a dict with a field named `color` containing `[f64; 3]` where the accent-color property is a `[f64; 3]`. So what I would do here instead, is to deserialize the property a Vec<f64>` instead and create a `Color` instance from it before returning the value

src/desktop/settings.rs
A6GibKm: Sorry, i didn't get that the `Item = Color` was different in the diff, move along.

src/desktop/color.rs
wash2: Ok I've refactored a bit. I understand how changes to the definition could be a problem for screenshot's pick color.

src/desktop/color.rs
```diff
@@ -13,6 +13,10 @@ pub struct Color {
 }
 
 impl Color {
+    pub(crate) fn new(r: f64, g: f64, b: f64) -> Self {
+        Self { color: [r, g, b] }
+    }
```
bilelmoussaoui, line 18:
```suggestion
    pub(crate) fn from_vec(color: Vec<f64>) -> Self {
        debug_assert_eq!(color.len(), 3);
        Self {
            color: <[f64; 3]>::try_from(color).unwrap(),
        }
    }
```

src/desktop/settings.rs
```diff
@@ -205,8 +205,9 @@ impl<'a> Settings<'a> {
 
     /// Retrieves the system's preferred accent color
     pub async fn accent_color(&self) -> Result<Color, Error> {
-        self.read::<Color>(APPEARANCE_NAMESPACE, ACCENT_COLOR_SCHEME_KEY)
+        self.read::<(f64, f64, f64)>(APPEARANCE_NAMESPACE, ACCENT_COLOR_SCHEME_KEY)
             .await
+            .map(|(r, g, b)| Color::new(r, g, b))
```
bilelmoussaoui, line 210:
```suggestion
        self.read::<Vec<f64>>(APPEARANCE_NAMESPACE, ACCENT_COLOR_SCHEME_KEY)
            .await
            .map(Color::from_vec)
```

bilelmoussaoui (commented):
Otherwise looks good now, thanks a lot

src/desktop/color.rs
wash2: Thanks for your suggestions and for reviewing the changes. I really just don't understand though, how (ddd) is translated to [f64;3]. When I build a `zvariant::Value` from a `Vec<f64>`, it is converted to a `zvariant::Value::Array` with the signature `ad` instead of `(ddd)`.

bilelmoussaoui: Oh indeed, because arrays are structs in dbus/serde world... annoying

bilelmoussaoui (approved):
thanks

src/desktop/color.rs
wash2: Also, I think if the screenshot color is also supposed to be `(ddd)` in the dict value, that the `Color` type should be (f64, f64, f64) internally

PR #197, opened 2024-03-04
refactor: Color.color is a tuple

wash2:
I have not tested this, but looking at https://github.com/flatpak/xdg-desktop-portal/blob/04fdd3579abef01248997608e2abd919be94c1b1/data/org.freedesktop.impl.portal.Screenshot.xml#L90 the definition of `Color` might have the same issue  as for the accent-color when used in the screenshot portal.

### DIFF
--- a/src/desktop/color.rs
+++ b/src/desktop/color.rs
@@ -1,35 +1,32 @@
-use crate::{
-    zvariant::{self, DeserializeDict, Type},
-    Error,
-};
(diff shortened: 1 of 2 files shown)


bilelmoussaoui:
The current code works perfectly fine. It is tested in ashod demo already

bilelmoussaoui:
Thanks for your contributions

PR #198, opened 2024-03-04
color fixes 0.8

wash2:
Should the color fixes be applied to 0.7? I guess the change to `receive_accent_color_changed` is breaking, so I'm not sure.

### DIFF
--- a/src/desktop/color.rs
+++ b/src/desktop/color.rs
@@ -1,35 +1,32 @@
-use crate::{
-    zvariant::{self, DeserializeDict, Type},
-    Error,
-};
+use crate::zvariant::{self, DeserializeDict, Type};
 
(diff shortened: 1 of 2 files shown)


bilelmoussaoui:
> Should the color fixes be applied to 0.7? I guess the change to receive_accent_color_changed is breaking, so I'm not sure.

It is that is true, but the old implementation was just wrong. It would have never worked, so it is ok to break API here.

wash2:
Is failing CI blocking a merge?

bilelmoussaoui:
Yes, please target 0.8 branch. I won't be updating the 0.7 version anymore.

bilelmoussaoui:
Thanks a lot ! will be doing a fix release sometime in the future once more fixes are collected

PR #199, opened 2024-03-16
Fix deserialization in input_capture portal

### LINKED ISSUE #187: input_capture portal Signals are not working
The `receive_activate()` / `receive_deactivate()` Streams on the `input_capture` portal never yield any Result.

I did some digging and could narrow it down a bit:

`receive_activated` calls `signal("Activated")` which yields a Stream that filter-maps on the result of `msg.body::<Activated>().ok()`

So I tried to replicated the `body()` call manually and it always throws the following error:
```
called `Result::unwrap()` on an `Err` value: Variant(Message("invalid type: character `o`, expected `v`, `a` or `(`"))
```
Which then of course results in the signal getting filtered by the filter_map call.

This error in `body` is originated from `zvariant::from_slice_fds_for_dynamic_signature` -> `deserializer_for_signature` -> ... -> `deserialize_seq` where it ultimately fails in `zbus/zvariant/src/dbus/de.rs:347` because it does not expect an `o`.

feschber:
closes #187

### DIFF
--- a/src/desktop/input_capture.rs
+++ b/src/desktop/input_capture.rs
@@ -107,7 +107,7 @@ impl Deactivated {
     }
 }
 
-#[derive(Debug, Deserialize, Type)]
+#[derive(Debug, DeserializeDict, Type)]
 #[zvariant(signature = "dict")]
 
 struct ActivatedOptions {
--- a/src/desktop/input_capture.rs
+++ b/src/desktop/input_capture.rs
@@ -83,7 +83,7 @@ impl Disabled {
     }
 }
 
-#[derive(Debug, Deserialize, Type)]
+#[derive(Debug, DeserializeDict, Type)]
 #[zvariant(signature = "dict")]
 struct DeactivatedOptions {
     activation_id: u32,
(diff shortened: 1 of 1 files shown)


A6GibKm:
Thanks! Just for the record, could you please add a small description on how this was tested here?

feschber:
the example I linked in the issue correctly deserializes the activation token now

feschber:
https://github.com/feschber/lan-mouse/tree/libei-producer this works correctly now

bilelmoussaoui:
Oh nooo. That is really stupid from my side. Would you mind checking other types that uses signature = dict and see if they use the correct deserialization derive macro?

Thanks a lot

feschber:
> Oh nooo. That is really stupid from my side. Would you mind checking other types that uses signature = dict and see if they use the correct deserialization derive macro?
> 
> Thanks a lot

No problem, I could 100% see myself doing the same.
I did a quick grep and the rest seemed to look fine

bilelmoussaoui:
I will backport and do a release later today

Issue #200, opened 2024-03-19
Add persistence and session restore for the RemoteDesktop portal.


Adamskye:
Like with the ScreenCast portal, the RemoteDesktop portal allows for specifying persist mode and may return some restore data so that a program wouldn't need to prompt the user every time they want to use it. This should probably be implemented the same way that it has been for the ScreenCast portal.

Adamskye:
There is already a struct for PersistMode in screencast.rs, and I wasn't sure whether I should have reused this one or created a new one for remote desktop. I opted for the latter option, if that's okay.

bilelmoussaoui:
This is now handled but let us keep the issue open till we add it to the demo as well

bilelmoussaoui:
Fixed in https://github.com/bilelmoussaoui/ashpd/commit/47aa993a24c20c568ab535f7f95cb597494dcf46

PR #201, opened 2024-03-19
Add remote desktop persistence and session restore.

Adamskye:
Addresses issue #200.

### DIFF
--- a/src/desktop/mod.rs
+++ b/src/desktop/mod.rs
@@ -60,7 +60,7 @@ pub mod wallpaper;
 
 #[cfg_attr(feature = "glib", derive(glib::Enum))]
 #[cfg_attr(feature = "glib", enum_type(name = "AshpdPersistMode"))]
-#[derive(Default, Serialize_repr, PartialEq, Eq, Debug, Copy, Clone, Type)]
+#[derive(Default, serde_repr::Serialize_repr, PartialEq, Eq, Debug, Copy, Clone, zbus::zvariant::Type)]
 #[doc(alias = "XdpPersistMode")]
 #[repr(u32)]
 /// Persistence mode for a screencast or remote desktop session.
--- a/src/desktop/mod.rs
+++ b/src/desktop/mod.rs
@@ -58,7 +58,9 @@ pub mod wallpaper;
 
 #[cfg_attr(feature = "glib", derive(glib::Enum))]
 #[cfg_attr(feature = "glib", enum_type(name = "AshpdPersistMode"))]
-#[derive(Default, serde_repr::Serialize_repr, PartialEq, Eq, Debug, Copy, Clone, zbus::zvariant::Type)]
+#[derive(
+    Default, serde_repr::Serialize_repr, PartialEq, Eq, Debug, Copy, Clone, zbus::zvariant::Type,
+)]
 #[doc(alias = "XdpPersistMode")]
 #[repr(u32)]
 /// Persistence mode for a screencast or remote desktop session.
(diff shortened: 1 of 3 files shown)


src/desktop/remote_desktop.rs
```diff
@@ -138,6 +148,25 @@ pub enum Axis {
...
+    #[default]
+    /// Do not persist.
+    DoNot = 0,
+    #[doc(alias = "XDP_REMOTEDESKTOP_PERSIST_MODE_TRANSIENT")]
+    /// Persist while the application is running.
+    Application = 1,
+    #[doc(alias = "XDP_REMOTEDESKTOP_PERSIST_MODE_PERSISTENT")]
+    /// Persist until explicitly revoked.
+    ExplicitlyRevoked = 2,
+}
```
bilelmoussaoui, line 168: Not a fan of this duplication. Just copy the enum to src/desktop/mod.rs and re-export it from screencast (with a deprecated warning) to avoid breaking existing code

bilelmoussaoui (commented):
Otherwise lgtm, unfortunately this is an API break so we can't land it in 0.8.2. It will have to wait until 0.9

src/desktop/remote_desktop.rs
Adamskye: Okay, I moved PersistMode into src/desktop/mod.rs. Unfortunately, it seems you can't add deprecated warnings to re-exports https://github.com/rust-lang/rust/issues/30827.

src/desktop/screencast.rs
```diff
@@ -46,6 +49,7 @@ use serde::Deserialize;
 use serde_repr::{Deserialize_repr, Serialize_repr};
 use zbus::zvariant::{self, DeserializeDict, SerializeDict, Type, Value};
 
+pub use super::PersistMode;
```
bilelmoussaoui, line 52:
```suggestion
#[deprecated = "Please use `ashpd::desktop::PersistMode` instead"]
pub use super::PersistMode;
```

src/desktop/remote_desktop.rs
bilelmoussaoui: Oh really, that is unfortunate.

bilelmoussaoui: Then, just remove the re-export as we can't land this in 0.8.x anyways.

src/desktop/mod.rs
```diff
@@ -10,6 +10,8 @@ mod color;
 pub use color::Color;
 mod icon;
 pub use icon::Icon;
+use serde_repr::Serialize_repr;
+use zbus::zvariant::Type;
```
bilelmoussaoui, line 14:
```suggestion
```

src/desktop/mod.rs
```diff
@@ -55,3 +57,22 @@ pub mod secret;
 pub mod settings;
 pub mod trash;
 pub mod wallpaper;
+
+#[cfg_attr(feature = "glib", derive(glib::Enum))]
+#[cfg_attr(feature = "glib", enum_type(name = "AshpdPersistMode"))]
+#[derive(Default, Serialize_repr, PartialEq, Eq, Debug, Copy, Clone, Type)]
```
bilelmoussaoui, line 63:
```suggestion
#[derive(Default, serde_repr::Serialize_repr, PartialEq, Eq, Debug, Copy, Clone, zbus::zvariant::Type)]
```

Instead

Adamskye:
I should probably update the demos too. I'll do this tomorrow.

bilelmoussaoui:
> I should probably update the demos too. I'll do this tomorrow.

The demo targets stable ashpd, so we will have to land this PR first. Then you can open a new one for the demo that switches the ashpd dependency to use ashpd from git instead. The said PR won't be merged until we release 0.9 in the future.

bilelmoussaoui:
Thanks for your contribution

PR #205, opened 2024-04-23
Misc gtk4 fixes

hfiguiere:
There is no reason to require Gtk 4.14 but to prevent building on Fedora 39 (or other recent distros).

Also fix a warning: 

```
(ashpd-demo:1693266): Adwaita-WARNING **: 20:36:20.845: AdwNavigationPage 0x55c88967d080 is missing a title. To hide a header bar title, consider using AdwHeaderBar:show-title instead.
```

### DIFF
--- a/ashpd-demo/Cargo.toml
+++ b/ashpd-demo/Cargo.toml
@@ -12,8 +12,8 @@ chrono = {version = "0.4", default-features = false, features = ["clock"]}
 futures-util = "0.3"
 gettext-rs = {version = "0.7", features = ["gettext-system"]}
 gst = {package = "gstreamer", version = "0.22"}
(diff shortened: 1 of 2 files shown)


bilelmoussaoui:
> There is no reason to require Gtk 4.14 but to prevent building on Fedora 39 (or other recent distros).

I actually intentionally used v4_14 there. The app is only shipped as a flatpak so that justification is not really enough for me to merge that change.

For the second commit, the fix is wrong. A proper title must be set instead of an empty string

hfiguiere:
It prevent contribution to require 4.14. Whether it's shipped as flatpak or not. as it is not necessary.

bilelmoussaoui:
> It prevent contribution to require 4.14. Whether it's shipped as flatpak or not. as it is not necessary.

There is a difference in the behaviour of the GStreamer plugin which is the whole reason why the feature was enabled.

Issue #208, opened 2024-05-19
Response deserialization broken on master branch


kb-1000:
I've been trying out the master branch because of #201, but unfortunately deserializing the Response signal contents is broken there:

```
◀ method_call: CreateSession [serial = 20]
  path = /org/freedesktop/portal/desktop
  sender = :1.393
  destination = org.freedesktop.portal.Desktop
  interface = org.freedesktop.portal.RemoteDesktop
  Data (a{sv}): (
       {'handle_token': ('s', 'ashpd_I0STYhaO0z'),
        'session_handle_token': ('s', 'ashpd_qvxZHxJk6z')},,
    )

▷ method_return:  [serial = 12403]
  Flags: 1 (no_reply_expected)
  destination = :1.393
  reply_serial = 20
  sender = :1.16
  Data (o): ('/org/freedesktop/portal/desktop/request/1_393/ashpd_I0STYhaO0z',)

▷ signal: Response [serial = 12404]
  Flags: 1 (no_reply_expected)
  path = /org/freedesktop/portal/desktop/request/1_393/ashpd_I0STYhaO0z
  interface = org.freedesktop.portal.Request
  destination = :1.393
  sender = :1.16
  Data (ua{sv}): (
       0,
       {'session_handle': ('s',
                           '/org/freedesktop/portal/desktop/session/1_393/ashpd_qvxZHxJk6z')},
    )

2024-05-19T18:44:59.575343Z  INFO ashpd::desktop::request: Received signal 'Response' on 'org.freedesktop.portal.Request'
Error: ZBus Error: invalid type: character `u`, expected `v`, `a` or `(`
```
It was apparently broken by https://github.com/bilelmoussaoui/ashpd/commit/7e754fb8ba601c635e7395e47aaefb28a836de54, if I revert it locally, everything works fine.

bilelmoussaoui:
Seems like a zbus bug to me to be honest as the signature is `ua{sv}` and not `(ua{sv})`

The funny thing is that 
```rust
        assert_eq!(
            <(ResponseType, HashMap<&str, Value<'_>>)>::signature(),
            Response::<()>::signature(),
        );
```
Passes, so the changes mentioned in the commit shouldn't affect things that much but somehow it does...

bilelmoussaoui:
See https://github.com/bilelmoussaoui/ashpd/commit/595a9655f69164286d4e0074b908baabf1540469

PR #210, opened 2024-05-20
USB Portal

hfiguiere:
This is a draft, needs a lot of cleanup. Also the portal hasn't landed yet.

But I'm using this to exercise / test the USB portal.


hfiguiere:
Currently includes a revert for #208

hfiguiere:
The build failure is because it needs a flatpak version that isn't available.

src/desktop/usb.rs
```diff
@@ -0,0 +1,259 @@
...
+        let session: OwnedObjectPath = self.0.call("CreateSession", &(&options)).await?;
+        Session::new(session).await
+    }
+
+    /// Enumerate USB devices.
+    ///
+    #[doc(alias = "EnumerateDevices")]
+    pub async fn enumerate_devices(
+        &self,
+        options: UsbOptions,
```
bilelmoussaoui, line 203: There is no options here, so just pass Default? Also seems to have both a request_handle, yet it doesn't use the mechanism for requests and do a plain dbus call?

src/desktop/usb.rs
```diff
@@ -0,0 +1,259 @@
...
+    }
+
+    /// Acquire devices
+    ///
+    /// The portal will perform the permission request.
+    ///
+    #[doc(alias = "AcquireDevices")]
+    pub async fn acquire_devices(
+        &self,
+        parent_window: &WindowIdentifier,
```
bilelmoussaoui, line 215: This should become `Option<&WindowIdentifier>` with the changes in main.

src/desktop/usb.rs
```diff
@@ -0,0 +1,259 @@
...
+        devices: &[(&String, bool)],
```
bilelmoussaoui, line 216: What are those string/bool supposed to be? I haven't looked at the spec, but it would be nice to clarify their meaning. If they are returned by some other method call, then they should have an alias to make it easier to use.

src/desktop/usb.rs
```diff
@@ -0,0 +1,259 @@
...
+#[derive(SerializeDict, Type, Debug, Default)]
+#[zvariant(signature = "dict")]
+pub struct AcquireDevice {
+    writable: bool,
+}
+
+/// Option for AcquireDevice call.
+#[derive(SerializeDict, Type, Debug, Default)]
+#[zvariant(signature = "dict")]
+pub struct AcquireOptions {
```
bilelmoussaoui, line 109: There is no reason for making this struct public

src/desktop/usb.rs
bilelmoussaoui: Ah I see from the rest of the code, imho, make it take a new type called `Device(String /* ID */ , bool / * writable */); or so.

src/desktop/usb.rs
```diff
@@ -0,0 +1,259 @@
...
+            .empty_request(
+                &options.handle_token,
+                "AcquireDevices",
+                &(parent_window, &acquire_devices, &options),
+            )
+            .await
+    }
+
+    /// Call on success of acquire_devices.
+    #[doc(alias = "FinishAcquireDevices")]
```
bilelmoussaoui, line 236: This seems like it can be just called at the end of acquire_device, why it is a separate method?

src/desktop/usb.rs
```diff
@@ -0,0 +1,259 @@
...
+#[derive(SerializeDict, Type, Debug, Default)]
+#[zvariant(signature = "dict")]
+pub struct AcquireOptions {
+    handle_token: HandleToken,
+}
+
+/// Finished device acquired.
+#[derive(DeserializeDict, Type, Debug, Default)]
+#[zvariant(signature = "dict")]
+pub struct AcquiredDevice {
```
bilelmoussaoui, line 116: This seems like it should be an internal type that gets mapped into a `Result<OwnedFd, UsbError>`, as it is, it is not really Rust friendly.

src/desktop/usb.rs
```diff
@@ -0,0 +1,259 @@
...
+use futures_util::Stream;
+use serde::Deserialize;
+use zbus::zvariant::{
+    DeserializeDict, Fd, ObjectPath, OwnedFd, OwnedObjectPath, OwnedValue, SerializeDict, Type,
+    Value,
+};
+
+use super::Request;
+use crate::{
+    desktop::{HandleToken, Session, SessionPortal},
```
bilelmoussaoui, line 42: add Request here, remove the user super::Request above

ashpd-demo/data/resources/ui/usbdevicerow.ui
```diff
@@ -0,0 +1,26 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<interface>
+  <template class="DeviceRow" parent="AdwActionRow">
```
bilelmoussaoui, line 3: Should be called UsbDeviceRow and the template usb_device_row or so

ashpd-demo/src/portals/desktop/usb.rs
```diff
@@ -0,0 +1,362 @@
+// Copyright (C) 2024 GNOME Foundation
+//
+// Authors:
+//     Hubert Figuière <hub@figuiere.net>
+//
+
+use std::cell::RefCell;
+use std::collections::HashMap;
```
bilelmoussaoui, line 8: We use rustfmt from rust nightly, it groups imports which makes it nicer

ashpd-demo/src/portals/desktop/usb.rs
```diff
@@ -0,0 +1,362 @@
...
+
+use crate::widgets::{PortalPage, PortalPageExt, PortalPageImpl};
+
+glib::wrapper! {
+    pub struct DeviceRow(ObjectSubclass<imp::DeviceRow>)
+        @extends gtk::Widget, gtk::ListBoxRow, adw::PreferencesRow, adw::ActionRow;
+}
+
+impl DeviceRow {
+    fn new() -> Self {
```
bilelmoussaoui, line 34: Impl default instead

ashpd-demo/src/portals/desktop/usb.rs
```diff
@@ -0,0 +1,362 @@
...
+
+    fn acquire(&self) {
+        self.imp().checkbox.set_active(true);
+    }
+
+    fn release(&self) {
+        self.imp().checkbox.set_active(false);
+    }
+
+    fn connect_share_clicked<F: Fn(&gtk::Button) + 'static>(&self, f: F) -> glib::SignalHandlerId {
```
bilelmoussaoui, line 46: Use template callbacks?

ashpd-demo/src/portals/desktop/usb.rs
```diff
@@ -0,0 +1,362 @@
...
+                    page,
+                    move |row| {
+                        glib::spawn_future_local(clone!(
+                            #[strong]
+                            row,
+                            #[strong]
+                            device_id,
+                            #[strong]
+                            page,
+                            async move {
```
bilelmoussaoui, line 173: Move to a separate function, maybe even inside DeviceRow as an async helper function, because as it is, the function is too huge

ashpd-demo/src/portals/desktop/usb.rs
```diff
@@ -0,0 +1,362 @@
...
+                                let root = row.native().unwrap();
+                                let identifier = WindowIdentifier::from_native(&root).await;
+                                let usb = UsbProxy::new().await.unwrap();
```
bilelmoussaoui: The unwraps here are not good, once you move it to a separate function, you can make it report errors using the notifications support in PortalPage

ashpd-demo/src/portals/desktop/usb.rs
```diff
@@ -0,0 +1,362 @@
...
+                    #[strong]
+                    page,
+                    move |_row| {
+                        glib::spawn_future_local(clone!(
+                            #[strong]
+                            device_id,
+                            #[strong]
+                            page,
+                            async move {
+                                let usb = UsbProxy::new().await.unwrap();
```
bilelmoussaoui: Keep the proxy around

ashpd-demo/src/portals/desktop/usb.rs
```diff
@@ -0,0 +1,362 @@
...
+                                        tracing::error!("Acquire device error: {err}");
+                                        page.error(&format!("Acquire device error: {err}"));
+                                    }
+                                }
+                            }
+                        ));
+                    }
+                ));
+                let device_id = device.0.clone();
+                row.connect_unshare_clicked(clone!(
```
bilelmoussaoui, line 180: Same, move it to a separate function

ashpd-demo/src/portals/desktop/usb.rs
```diff
@@ -0,0 +1,362 @@
...
+            Ok(())
+        }
+
+        pub(super) async fn start_session(&self) -> ashpd::Result<()> {
+            let usb = UsbProxy::new().await?;
+            let session = usb.create_session().await?;
+            self.session.lock().await.replace(session);
+
+            let session = self.session.clone();
+            glib::spawn_future(async move {
```
bilelmoussaoui, line 267: It is already async, no need for using glib here

ashpd-demo/src/portals/desktop/usb.rs
```diff
@@ -0,0 +1,362 @@
...
+                tracing::debug!("Loop is gone");
+                Ok::<(), ashpd::Error>(())
+            });
+
+            Ok(())
+        }
+
+        pub(super) async fn stop_session(&self) -> anyhow::Result<()> {
+            if let Some(session) = self.session.lock().await.take() {
+                session.close().await?;
```
bilelmoussaoui, line 294: should switch the buttons state

src/desktop/usb.rs
hfiguiere: the problem is that there is a dbus limitation one how many fd it can return, so we need to be able to chunk the finish call. So it has to be called manually.

ashpd-demo/src/portals/desktop/usb.rs
hfiguiere: this is done by the non-imp version which is the one called from the UI.

src/desktop/usb.rs
bilelmoussaoui: We talked about this in the matrix channel, this is a dbus limitation. You should instead call it yourself in AcquireDevices until it stops returning, and return the results or even better make it return a Stream so you can call next on it which would call internally the FinishAcquireDevices.

hfiguiere: This is now a private method and `AcquireDevices` will return the devices directly.

ashpd-demo/src/portals/desktop/usb.rs
hfiguiere: done

ashpd-demo/build-aux/com.belmoussaoui.ashpd.demo.Devel.json
```diff
@@ -30,6 +32,16 @@
         ]
     },
     "modules": [
+        {
+            "name": "libusb",
```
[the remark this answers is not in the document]
hfiguiere: if that was a flathub submission, definitely. But here, I don't see any benefit.

ashpd-demo/data/resources/ui/usb_device_row.ui
```diff
@@ -0,0 +1,29 @@
...
+      <object class="GtkCheckButton" id="checkbox">
+        <property name="sensitive">false</property>
+      </object>
+    </child>
+    <child type="suffix">
+      <object class="GtkBox" id="box1">
+        <property name="margin-top">6</property>
+        <property name="margin-bottom">6</property>
+        <style>
+          <class name="linked" />
```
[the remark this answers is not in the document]
hfiguiere: link (to the documentation) ?

ashpd-demo/src/portals/desktop/usb.rs
```diff
@@ -0,0 +1,370 @@
...
+                    for ev in events {
+                        println!(
+                            "Received event: {} for device {}",
+                            ev.event_action(),
+                            ev.event_device_id()
+                        );
+                    }
+                }
+            }
+            tracing::debug!("Loop is gone");
```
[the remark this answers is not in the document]
hfiguiere: well there is still some cleanup needed. DRAFT.

ashpd-demo/data/resources/ui/usb_device_row.ui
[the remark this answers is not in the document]
A6GibKm: There is none, ask in the #design:gnome.org room if you would like more context.

src/desktop/usb.rs
```diff
@@ -0,0 +1,281 @@
...
+#[zvariant(signature = "dict")]
+struct UsbEnumerateOptions {}
+
+/// USB device description
+#[derive(SerializeDict, DeserializeDict, Type, Debug, Default)]
+#[zvariant(signature = "dict")]
+pub struct UsbDevice {
+    parent: Option<String>,
+    readable: Option<bool>,
+    /// Device is writable. Default is false.
```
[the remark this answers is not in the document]
hfiguiere: "Default is false". I thought this was clear.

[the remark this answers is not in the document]
hfiguiere: documented `readable`. (cf the D-Bus xml)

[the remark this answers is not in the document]
A6GibKm: What does `None` mean? How is it different from `false`? Why would you get one over the other?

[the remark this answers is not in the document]
hfiguiere: because it's the `a{sv}` from D-Bus. As per the D-Bus interface it can be omitted so it has to be omittable here.

ashpd-demo/build-aux/com.belmoussaoui.ashpd.demo.Devel.json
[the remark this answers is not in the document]
bilelmoussaoui: Agree with @hfiguiere

ashpd-demo/build-aux/com.belmoussaoui.ashpd.demo.Devel.json
```diff
@@ -50,14 +62,15 @@
         {
             "name": "ashpd-demo",
```
bilelmoussaoui, line 63: I rather have a change that puts the manifest in ./ and see if that helps? :)

ashpd-demo/src/portals/desktop/usb.rs
```diff
@@ -0,0 +1,373 @@
...
+    fn acquire(&self) {
+        self.imp().checkbox.set_active(true);
+    }
+
+    fn release(&self) {
+        self.imp().checkbox.set_active(false);
+    }
+}
+
+mod imp {
```
bilelmoussaoui, line 50: This doesn't follow the current code style that puts mod imp at the top of the file

ashpd-demo/src/portals/desktop/usb.rs
```diff
@@ -0,0 +1,373 @@
...
+    }
+
+    impl ObjectImpl for UsbDeviceRow {}
+    impl WidgetImpl for UsbDeviceRow {}
+    impl ListBoxRowImpl for UsbDeviceRow {}
+    impl PreferencesRowImpl for UsbDeviceRow {}
+    impl ActionRowImpl for UsbDeviceRow {}
+
+    #[derive(Debug, gtk::CompositeTemplate, Default)]
+    #[template(resource = "/com/belmoussaoui/ashpd/demo/usb.ui")]
```
bilelmoussaoui, line 118: and split different types in a separate mod, see https://github.com/bilelmoussaoui/ashpd/blob/master/ashpd-demo/src/portals/desktop/notification.rs#L161

src/desktop/request.rs
```diff
@@ -268,7 +268,8 @@ where
         self.0.call("Close", &()).await
     }
 
-    pub(crate) fn path(&self) -> &ObjectPath<'_> {
+    /// Return the object path of the request.
+    pub fn path(&self) -> &ObjectPath<'_> {
```
bilelmoussaoui, line 272: Why was this made public?

src/desktop/usb.rs
```diff
@@ -0,0 +1,282 @@
...
+    /// Device can be opened for reading. Default is false.
+    readable: Option<bool>,
+    /// Device can be opened for writing. Default is false.
+    pub writable: Option<bool>,
+    /// The device node for the USB.
+    #[zvariant(rename = "device-file")]
+    pub device_file: Option<String>,
+    /// Device properties
+    pub properties: Option<HashMap<String, OwnedValue>>,
+}
```
bilelmoussaoui, line 71: The way we deal with structs in general is that no field is public, only add getters or setters that would be public. This allow us to change the way we deserialize/serialize things without impacting the public API.

src/desktop/usb.rs
```diff
@@ -0,0 +1,282 @@
...
+struct AcquireOptions {
+    handle_token: HandleToken,
+}
+
+/// Finished device acquired.
+#[derive(DeserializeDict, Type, Debug, Default)]
+#[zvariant(signature = "dict")]
+struct AcquiredDevice {
+    success: bool,
+    fd: Option<OwnedFd>,
```
bilelmoussaoui, line 135: So if success = true, do we always receive an fd or not? or if the success = true and fd = None, there are no more fds to acquire?

src/desktop/usb.rs
```diff
@@ -0,0 +1,282 @@
...
+    error: Option<String>,
+}
+
+#[derive(Debug, Deserialize, Type)]
+/// A usb event received part of the `device_event` signal response.
+pub struct UsbEvent(/* action */ String, /* id */ String, UsbDevice);
```
bilelmoussaoui, line 141: Needs getters.

src/desktop/usb.rs
```diff
@@ -0,0 +1,282 @@
...
+        self.0.as_ref()
+    }
+
+    /// Events received
+    pub fn events(&self) -> &[UsbEvent] {
+        &self.1
+    }
+}
+
+impl UsbEvent {
```
bilelmoussaoui, line 159: Ah this should go up

src/desktop/usb.rs
```diff
@@ -0,0 +1,282 @@
...
+            }
+        }
+        Ok(devices)
+    }
+
+    /// Call on success of acquire_devices.
+    #[doc(alias = "FinishAcquireDevices")]
+    async fn finish_acquire_devices(
+        &self,
+        handle_token: &ObjectPath<'_>,
```
bilelmoussaoui, line 248: Why is this param called handle_token? is it really one or a request path?

src/desktop/usb.rs
```diff
@@ -0,0 +1,282 @@
...
+pub struct UsbProxy<'a>(Proxy<'a>);
+
+impl<'a> UsbProxy<'a> {
+    /// Create a new instance of [`UsbProxy`].
+    pub async fn new() -> Result<UsbProxy<'a>, Error> {
+        let proxy = Proxy::new_desktop("org.freedesktop.portal.Usb").await?;
+        Ok(Self(proxy))
+    }
+
+    /// Create Session
```
bilelmoussaoui, line 188: Please follow the doc style for proxies that are public.

src/desktop/usb.rs
```diff
@@ -0,0 +1,282 @@
...
+                &(&parent_window, &acquire_devices, &options),
+            )
+            .await?;
+        let mut devices: Vec<(String, Result<OwnedFd, UsbError>)> = vec![];
+        if request.response().is_ok() {
+            let path = request.path();
+            loop {
+                let mut result = self.finish_acquire_devices(path).await?;
+                let finished = result.1;
+                devices.append(&mut result.0);
```
bilelmoussaoui, line 235:
```suggestion
                let (new_devices, finished) = self.finish_acquire_devices(path).await?;
                devices.append(&new_devices);
```
?

src/desktop/usb.rs
```diff
@@ -0,0 +1,282 @@
...
+
+/// Device to acquire.
+#[derive(SerializeDict, Type, Debug, Default)]
+#[zvariant(signature = "dict")]
+pub struct AcquireDevice {
+    writable: bool,
+}
+
+/// Device to acquire
+pub struct Device(pub String /* ID */, pub bool /* writable */);
```
bilelmoussaoui, line 121: This needs a constructor/getters

src/desktop/usb.rs
```diff
@@ -0,0 +1,282 @@
...
+                .get("ID_MODEL_FROM_DATABASE")
+                .or_else(|| properties.get("ID_MODEL_ENC"))
+                .and_then(|v| v.downcast_ref::<String>().ok())
+        })
+    }
+}
+
+/// USB error for acquiring device.
+#[derive(Debug)]
+pub struct UsbError(pub Option<String>);
```
bilelmoussaoui, line 97: Needs to implement std::error::Error

src/desktop/usb.rs
hfiguiere: If success is true we expect the fd, otherwise the error is set.

This doesn't preclude other devices to be successfully acquired.

src/desktop/request.rs
hfiguiere: was a mistake likely.

src/desktop/usb.rs
hfiguiere: adapted. (new_devices needs to be mutable)

src/desktop/usb.rs
hfiguiere: the getters have been added but aren't used. Maybe I should just omit them as the struct is only constructed (added a `new`)

src/desktop/usb.rs
hfiguiere: changed.

src/desktop/usb.rs
```diff
@@ -0,0 +1,326 @@
...
+    /// The device node for the USB.
+    #[zvariant(rename = "device-file")]
+    device_file: Option<String>,
+    /// Device properties
+    properties: Option<HashMap<String, OwnedValue>>,
+}
+
+impl UsbDevice {
+    /// Return if the device is writable.
+    pub fn writable(&self) -> bool {
```
bilelmoussaoui, line 75:
```suggestion
    pub fn is_writable(&self) -> bool {
```

src/desktop/usb.rs
```diff
@@ -0,0 +1,326 @@
...
+        self.writable.unwrap_or(false)
+    }
+
+    /// Return the optional device file.
+    pub fn device_file(&self) -> Option<&String> {
+        self.device_file.as_ref()
+    }
```
bilelmoussaoui, line 82:
```suggestion
    pub fn device_file(&self) -> Option<&str> {
        self.device_file.as_deref()
    }
```

src/desktop/usb.rs
```diff
@@ -0,0 +1,326 @@
...
+/// Options for the USB portal.
+#[derive(SerializeDict, Type, Debug, Default)]
+#[zvariant(signature = "dict")]
+struct UsbEnumerateOptions {}
+
+/// USB device description
+#[derive(SerializeDict, DeserializeDict, Type, Debug, Default)]
+#[zvariant(signature = "dict")]
+pub struct UsbDevice {
+    parent: Option<String>,
```
bilelmoussaoui, line 61: this lacks getters for parent/readable/properties

src/desktop/usb.rs
```diff
@@ -0,0 +1,326 @@
...
+impl std::fmt::Display for UsbError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}", self.0.as_deref().unwrap_or(""))
+    }
+}
+
+impl std::error::Error for UsbError {
+    fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
+        None
+    }
```
bilelmoussaoui, line 118:
```suggestion
```

src/desktop/usb.rs
```diff
@@ -0,0 +1,326 @@
...
+        Device(id, writable)
+    }
+
+    /// Get the device id.
+    pub fn id(&self) -> &str {
+        &self.0
+    }
+
+    /// Get if the device is writable.
+    pub fn writable(&self) -> bool {
```
bilelmoussaoui, line 157:
```suggestion
    pub fn is_writable(&self) -> bool {
```

bilelmoussaoui (commented):
looks much better

src/desktop/usb.rs
```diff
@@ -0,0 +1,326 @@
...
+    /// See also [`CreateSession`](https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.portal.Usb.html#org-freedesktop-portal-usb-createsession).
+    #[doc(alias = "CreateSession")]
+    pub async fn create_session(&self) -> Result<Session<'a, Self>, Error> {
+        let options = CreateSessionOptions::default();
+        let session: OwnedObjectPath = self.0.call("CreateSession", &(&options)).await?;
+        Session::new(session).await
+    }
+
+    /// Enumerate USB devices.
+    #[doc(alias = "EnumerateDevices")]
```
bilelmoussaoui, line 242: This and the below methods lack links to the upstream docs, similar to what you have in create_session.

It would also be nice to describe what is returned in the various functions.

bilelmoussaoui:
thanks!

PR #217, opened 2024-06-19
session: Tie up portals to their corresponding sessions

### LINKED ISSUE #216: Figure out a way to tie a Session to a specific portal
Currently, you can create a Session object in any of the portals that have a `create_session` method and pass it to another portal which is not expected.

The current portals that can create a Session nowadays:
- Location
- ScreenCast
- RemoteDesktop
- InputCapture
- GlobalShortcuts

The problem here, is that you could pass a session created with RemoteDesktop to some ScreenCast protal calls like [Screencast::select_sources()](https://bilelmoussaoui.github.io/ashpd/ashpd/desktop/screencast/struct.Screencast.html#method.select_sources) or any of the Clipboard portal calls.

Currently, the Session object is defined as `pub struct Session<'a>(Proxy<'a>);`

I guess, one could create some sort of `SessionFlags` bitfield and attach it to the session. Then any method that takes a Session could validate the passed session, not sure if there is any way to do that at compile time instead of runtime.

bilelmoussaoui:
Fixes #216

### DIFF
--- a/src/desktop/screencast.rs
+++ b/src/desktop/screencast.rs
@@ -49,7 +49,10 @@ use serde::Deserialize;
 use serde_repr::{Deserialize_repr, Serialize_repr};
 use zbus::zvariant::{self, DeserializeDict, SerializeDict, Type, Value};
 
-use super::{HandleToken, PersistMode, Request, Session};
+use super::{
+    remote_desktop::RemoteDesktop, session::SessionPortal, HandleToken, PersistMode, Request,
+    Session,
+};
 use crate::{desktop::session::CreateSessionResponse, proxy::Proxy, Error, WindowIdentifier};
 
 #[bitflags]
@@ -274,7 +277,7 @@ impl<'a> Screencast<'a> {
     /// See also [`CreateSession`](https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.portal.ScreenCast.html#org-freedesktop-portal-screencast-createsession).
     #[doc(alias = "CreateSession")]
     #[doc(alias = "xdp_portal_create_screencast_session")]
-    pub async fn create_session(&self) -> Result<Session<'a>, Error> {
+    pub async fn create_session(&self) -> Result<Session<'a, Self>, Error> {
         let options = CreateSessionOptions::default();
         let (request, proxy) = futures_util::try_join!(
             self.0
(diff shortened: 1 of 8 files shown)


src/desktop/screencast.rs
```diff
@@ -417,3 +423,11 @@ impl<'a> std::ops::Deref for Screencast<'a> {
...
     }
 }
+
+impl SessionPortal for Screencast<'_> {}
+
+/// Defines which portals session can be used in a screen-cast.
+pub trait ScreencastSession: Serialize + zvariant::Type + Debug {}
+
+impl ScreencastSession for Session<'_, Screencast<'_>> {}
+impl ScreencastSession for Session<'_, RemoteDesktop<'_>> {}
```
sdroege, line 433:
```suggestion
pub trait HasScreencastSession {}

impl HasScreencastSession for Screencast<'_> {}
impl HasScreencastSession for RemoteDesktop<'_> {}
```

and then use `Session<'_, impl HasScreencastSession>` for the parameters. That way it's guaranteed to be a session and you don't need all the trait bounds

sdroege:
```diff
diff --git a/src/desktop/screencast.rs b/src/desktop/screencast.rs
index e427aa1..ae2eda3 100644
--- a/src/desktop/screencast.rs
+++ b/src/desktop/screencast.rs
@@ -45,7 +45,7 @@ use std::{collections::HashMap, fmt::Debug, os::fd::OwnedFd};
 
 use enumflags2::{bitflags, BitFlags};
 use futures_util::TryFutureExt;
-use serde::{Deserialize, Serialize};
+use serde::Deserialize;
 use serde_repr::{Deserialize_repr, Serialize_repr};
 use zbus::zvariant::{self, DeserializeDict, SerializeDict, Type, Value};
 
@@ -307,7 +307,7 @@ impl<'a> Screencast<'a> {
     #[doc(alias = "OpenPipeWireRemote")]
     pub async fn open_pipe_wire_remote(
         &self,
-        session: &impl ScreencastSession,
+        session: &Session<'_, impl HasScreencastSession>,
     ) -> Result<OwnedFd, Error> {
         // `options` parameter doesn't seems to be used yet
         // see https://github.com/flatpak/xdg-desktop-portal/blob/master/src/screen-cast.c#L812
@@ -341,7 +341,7 @@ impl<'a> Screencast<'a> {
     #[doc(alias = "SelectSources")]
     pub async fn select_sources(
         &self,
-        session: &impl ScreencastSession,
+        session: &Session<'_, impl HasScreencastSession>,
         cursor_mode: CursorMode,
         types: BitFlags<SourceType>,
         multiple: bool,
@@ -382,7 +382,7 @@ impl<'a> Screencast<'a> {
     #[doc(alias = "Start")]
     pub async fn start(
         &self,
-        session: &impl ScreencastSession,
+        session: &Session<'_, impl HasScreencastSession>,
         identifier: &WindowIdentifier,
     ) -> Result<Request<Streams>, Error> {
         let options = StartCastOptions::default();
@@ -427,7 +427,7 @@ impl<'a> std::ops::Deref for Screencast<'a> {
 impl SessionPortal for Screencast<'_> {}
 
 /// Defines which portals session can be used in a screen-cast.
-pub trait ScreencastSession: Serialize + zvariant::Type + Debug {}
+pub trait HasScreencastSession: SessionPortal {}
 
-impl ScreencastSession for Session<'_, Screencast<'_>> {}
-impl ScreencastSession for Session<'_, RemoteDesktop<'_>> {}
+impl HasScreencastSession for Screencast<'_> {}
+impl HasScreencastSession for RemoteDesktop<'_> {}
```

PR #218, opened 2024-06-19
make activation-options optional

feschber:
The spec states "options - Vardict with **_optional_** further information".

As of right now, the KDE portal does not send a `barrier_id` in the Activated signal, which breaks deserialization...

I'm proposing this change, which makes all of the fields optional.

Technically this implementation is also wrong with most other signals, but let me know, what you think about this.

### DIFF
--- a/src/desktop/input_capture.rs
+++ b/src/desktop/input_capture.rs
@@ -402,9 +402,9 @@ impl Deactivated {
 #[derive(Debug, DeserializeDict, Type)]
 #[zvariant(signature = "dict")]
 struct ActivatedOptions {
-    activation_id: u32,
-    cursor_position: (f32, f32),
-    barrier_id: BarrierID,
+    activation_id: Option<u32>,
+    cursor_position: Option<(f32, f32)>,
+    barrier_id: Option<BarrierID>,
 }
 
 /// Indicates that an input capturing session was activated.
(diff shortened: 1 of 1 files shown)


feschber:
Tested on https://github.com/feschber/ashpd-mre

bilelmoussaoui:
Optional as in more stuff could be added in future versions of the portal, not that all current ones shouldn't be sent... but then i have no idea about input capture in general. I wonder if KDE should instead the barrier_id here :)

feschber:
> I wonder if KDE should instead the barrier_id here :)

I agree. However I asked in the kwin matrix channel about this and the developer was of the opinion that the fields are optional. The spec is a bit ambiguous there I would say

bilelmoussaoui:
> > I wonder if KDE should instead the barrier_id here :)
> 
> I agree. However I asked in the kwin matrix channel about this and the developer was of the opinion that the fields are optional. The spec is a bit ambiguous there I would say

after a discussion in the xdg-portals matrix channel, it seems like they should indeed be optional. Although, I think before merging this, i would have a look at the various portals and see where we are not making things optional when they should be

bilelmoussaoui:
Seems like this is the only affected place.

bilelmoussaoui:
Please, in the future, prefix commits with `desktop/$portalName` to make it easier to find changes per portal

feschber:
The other signals in the input capture portal are affected as well, right?

feschber:
> Please, in the future, prefix commits with `desktop/$portalName` to make it easier to find changes per portal

Will do!

bilelmoussaoui:
> The other signals in the input capture portal are affected as well, right?

Right, I completely missed that this PR doesn't fix all of them, would you mind sending another PR for the rest?

PR #219, opened 2024-06-20
Allow setting activation_token in supported portals

apricotbucket28:
Allow manually setting the `activation_token` option for the [OpenURI](https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.portal.OpenURI.html) and [DynamicLauncher.Launch](https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.portal.DynamicLauncher.html#org-freedesktop-portal-dynamiclauncher-launch) portals.

I saw that https://github.com/bilelmoussaoui/ashpd/pull/70 is opened but it also includes additional integration, while this just allows setting the token if we already got it from somewhere else (e.g. directly via the Wayland protocol)

(related commit: https://github.com/apricotbucket28/zed/commit/15489a03966b1c627c9596e0635c5f45c9411b9e)

### DIFF
--- a/src/activation_token/mod.rs
+++ b/src/activation_token/mod.rs
@@ -1,0 +1,48 @@
+use std::ops::Deref;
+
+use serde::{Deserialize, Serialize};
+use zbus::zvariant::Type;
+
+/// A token that can be used to activate an application.
+///
+/// No guarantees are made for the token structure.
+#[derive(Debug, Deserialize, Serialize, Type, PartialEq, Eq, Hash, Clone)]
+pub struct ActivationToken(String);
+
+impl From<String> for ActivationToken {
+    fn from(value: String) -> Self {
+        Self(value)
+    }
+}
+
+impl From<&str> for ActivationToken {
+    fn from(value: &str) -> Self {
+        Self(value.to_owned())
+    }
+}
+
+impl From<ActivationToken> for String {
+    fn from(value: ActivationToken) -> String {
+        value.0
+    }
+}
+
+impl Deref for ActivationToken {
+    type Target = str;
+
+    fn deref(&self) -> &Self::Target {
+        self.0.as_str()
+    }
+}
+
+impl AsRef<str> for ActivationToken {
+    fn as_ref(&self) -> &str {
+        self.0.as_str()
+    }
+}
+
+impl std::fmt::Display for ActivationToken {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.write_str(self.as_ref())
+    }
+}
--- a/src/desktop/dynamic_launcher.rs
+++ b/src/desktop/dynamic_launcher.rs
@@ -58,7 +58,7 @@ use serde_repr::{Deserialize_repr, Serialize_repr};
 use zbus::zvariant::{self, DeserializeDict, OwnedValue, SerializeDict, Type, Value};
 
 use super::{HandleToken, Icon, Request};
-use crate::{proxy::Proxy, Error, WindowIdentifier};
+use crate::{proxy::Proxy, ActivationToken, Error, WindowIdentifier};
 
 #[bitflags]
 #[derive(Default, Serialize_repr, Deserialize_repr, PartialEq, Eq, Debug, Copy, Clone, Type)]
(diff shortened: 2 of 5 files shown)


src/desktop/dynamic_launcher.rs
```diff
@@ -321,9 +321,15 @@ impl<'a> DynamicLauncherProxy<'a> {
...
-        let options: HashMap<&str, zvariant::Value<'_>> = HashMap::new();
+    pub async fn launch(
+        &self,
+        desktop_file_id: &str,
+        activation_token: impl Into<Option<String>>,
+    ) -> Result<(), Error> {
+        let mut options: HashMap<&str, zvariant::Value<'_>> = HashMap::new();
+        if let Some(activation_token) = activation_token.into() {
+            options.insert("activation_token", activation_token.into());
+        }
```
bilelmoussaoui, line 332: This should be a struct similar to other options we have in the library.

bilelmoussaoui: But also I am not sure we want to expose the activation token as a random string, I would instead make it a new type that we could potentially validate

bilelmoussaoui: I know that you can't really validate an activation token under wayland but I don't know if it is the case for x11. Also, we would ideally expose in the future a way to create an activation token from a wayland / x11 surface, similar to how we do for the WindowIdentifier. See #70

bilelmoussaoui (commented):
While you are at it, we should also expose the activation_token in the email portal

src/desktop/dynamic_launcher.rs
apricotbucket28: Looking at the validation part, it seems like neither Wayland nor [X11](https://github.com/rust-windowing/winit/blob/db2c97a995bc403e0442675c7364010b806d5431/src/platform_impl/linux/x11/activation.rs#L19) have a unified standard for the token. Should I still add an `ActivationToken` type anyway?

bilelmoussaoui: I would still define a new type as we would need to provide constructors and so on for later. For the validation part, it would be nice to document that both x11 and wayland activation tokens can't be validated, so just implement From<String> and other traits.

src/desktop/email.rs
```diff
@@ -177,11 +176,9 @@ impl EmailRequest {
         self
     }
 
-    // TODO Added in version 4 of the interface.
```
bilelmoussaoui, line 180: Please keep this todo bit, as we would like to potentially only send stuff that are supported by the portal (micro-optimization, probably worthless but oh well)

src/desktop/open_uri.rs
```diff
@@ -171,6 +171,13 @@ impl OpenFileRequest {
         self
     }
 
+    /// Sets the token that can be used to activate the chosen application.
```
bilelmoussaoui, line 174: I see that you are only using this in your zed PR, please split it in a separate commit so that I could backport it for you otherwise you would have to wait for a major release in few weeks

apricotbucket28: Appreciate the concerns, but I think that Zed could switch to the git repo for now (Zed on Linux is still in development) :slightly_smiling_face:

src/activation_token/mod.rs
```diff
@@ -0,0 +1,42 @@
...
+///
+/// No guarantees are made for the token structure.
+#[derive(Debug, Deserialize, Serialize, Type, PartialEq, Eq, Hash, Clone)]
+pub struct ActivationToken(String);
+
+impl From<String> for ActivationToken {
+    fn from(value: String) -> Self {
+        ActivationToken(value)
+    }
+}
```
bilelmoussaoui, line 16: would you mind adding a `From<&str>` variant as well? for ease of use

src/activation_token/mod.rs
```diff
@@ -0,0 +1,42 @@
...
+
+/// A token that can be used to activate an application.
+///
+/// No guarantees are made for the token structure.
+#[derive(Debug, Deserialize, Serialize, Type, PartialEq, Eq, Hash, Clone)]
+pub struct ActivationToken(String);
+
+impl From<String> for ActivationToken {
+    fn from(value: String) -> Self {
+        ActivationToken(value)
```
bilelmoussaoui, line 14:
```suggestion
        Self(value)
```

bilelmoussaoui (approved):
otherwise lgtm

bilelmoussaoui:
Thanks a lot!

PR #220, opened 2024-06-22
StatusNotifier

XDeme1:
This is still a draft protocol, but its been in use for a while.
KDE has builtin support for tray icons, GNOME needs an extension to enable them, not sure about others.

Not sure if this is the right place for this.


bilelmoussaoui:
I am not sure why is this supposed to be part of ashpd?

PR #222, opened 2024-07-13
file_chooser: Make choices optional

apricotbucket28:
While investigating why the open directory dialogs in Zed weren't the same as other GTK programs and Chromium, I noticed that the dbus message ashpd sent had an empty choices array (even though Zed didn't set any choices itself). Making the choices optional so ashpd wouldn't send them fixed the problem.

This is probably a KDE quirk; but the [file chooser](https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.portal.FileChooser.html) portal documentation states:
> As a special case, passing an empty array for the list of choices indicates a boolean choice that is typically displayed as a check button, using “true” and “false” as the choices.

So it's probably a good idea to make them optional.

To clarify, this is what dialogs looked like before:
image

And this is how they look now:
image

### DIFF
--- a/src/desktop/file_chooser.rs
+++ b/src/desktop/file_chooser.rs
@@ -233,7 +233,7 @@ struct OpenFileOptions {
     directory: Option<bool>,
     filters: Vec<FileFilter>,
     current_filter: Option<FileFilter>,
-    choices: Vec<Choice>,
+    choices: Option<Vec<Choice>>,
     current_folder: Option<FilePath>,
 }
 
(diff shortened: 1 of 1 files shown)


bilelmoussaoui (approved):

bilelmoussaoui:
This is actually an annoying limitation of the SerializeDict macro that doesn't provide a way to skip a field if it is empty... For the time being, this looks good enough for me. Thanks!

PR #227, opened 2024-08-03
dynamic_laucher: Fix textview style

A6GibKm:
We need `.inline` to remove the `.view`'s background, but card and inline can't be on the same widget as card adds a background.

image

### DIFF
--- a/ashpd-demo/data/resources/ui/dynamic_launcher.ui
+++ b/ashpd-demo/data/resources/ui/dynamic_launcher.ui
@@ -76,11 +76,18 @@
                   </object>
                 </child>
                 <child>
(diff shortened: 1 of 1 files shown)


A6GibKm:
Note that TextView is scrollable and it is UB to have a parent that is not a ScrolledWindow. Using a scrolled window though requires the following hack:

```css
toolbarview scrolledwindow.card undershoot.top,
toolbarview scrolledwindow.card undershoot.bottom {
  box-shadow: none;
  background: none;
}
```

to hide separators added by the parent toolbarview.

bilelmoussaoui:
This one is no longer needed, right?

A6GibKm:
Whoops, no

PR #234, opened 2024-09-14
desktop/game_mode: Use `i32` for pid

kjarosh:
As per the documentation, methods which accept pids (e.g. `RegisterGame`, `RegisterGameByPid`) require the type of pid to be `i` (`i32`), not `u` (`u32`).

Using `u32` causes errors in runtime, making these methods unusable. This bug was most likely introduced by ae81250.

See https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.portal.GameMode.html

### DIFF
--- a/src/desktop/game_mode.rs
+++ b/src/desktop/game_mode.rs
@@ -20,7 +20,7 @@ use std::{fmt::Debug, os::fd::BorrowedFd};
 use serde_repr::Deserialize_repr;
 use zbus::zvariant::{Fd, Type};
 
-use crate::{error::PortalError, proxy::Proxy, Error};
+use crate::{error::PortalError, proxy::Proxy, Error, Pid};
 
 #[cfg_attr(feature = "glib", derive(glib::Enum))]
 #[cfg_attr(feature = "glib", enum_type(name = "AshpdGameModeStatus"))]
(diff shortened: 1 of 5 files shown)


src/desktop/game_mode.rs
```diff
@@ -154,7 +154,7 @@ impl<'a> GameMode<'a> {
     /// See also [`RegisterGame`](https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.portal.GameMode.html#org-freedesktop-portal-gamemode-registergame).
     #[doc(alias = "RegisterGame")]
     pub async fn register(&self, pid: u32) -> Result<(), Error> {
-        let status = self.0.call("RegisterGame", &(pid)).await?;
+        let status = self.0.call("RegisterGame", &(pid as i32)).await?;
```
bilelmoussaoui, line 157: why not replace the function signature? is this in order to have a backportable variant?

kjarosh: Originally, I had two reasons not to change the signature:
1. Backward compatibility, in order not to break code that already uses `u32`. (Now writing this I realized that probably no one uses it as it doesn't work...)
2. Better interoperability with Rust, as Rust uses `u32` for pids: https://doc.rust-lang.org/std/process/fn.id.html, https://doc.rust-lang.org/std/process/struct.Child.html#method.id

bilelmoussaoui: Right, in such case I would add 

```
// Matches the type used in std
pub type Pid = u32;
```

and pass that to the functions instead.

kjarosh: Sure thing, added the `Pid` type alias and replaced most pids in API. There are instances where `u64` is used though, but I haven't touched them.

Let me know if that's what you meant.

bilelmoussaoui (approved):
lgtm, thanks

torokati44:
Do you think we could get a release on `crates.io` with this fix sometime soon?

bilelmoussaoui:
> Do you think we could get a release on `crates.io` with this fix sometime soon?

sure, later this week

torokati44:
Great, thanks!

torokati44:
(a polite bump on this :no_mouth: )

bilelmoussaoui:
ah, sorry, completely forgot about it. let me do a release now

torokati44:
No worries, thanks a lot!

bilelmoussaoui:
0.9.2 should be up

torokati44:
Awesome, that was quick! ^^

PR #237, opened 2024-09-21
Shutdown unix streams before dropping

A6GibKm:


A6GibKm:
cc @SeaDve  this fixes the issue in Bustle 0.9.

SeaDve:
> cc @SeaDve this fixes the issue in Bustle 0.9.

Interesting. Still, there's probably an issue somewhere, causing bustle/zbus to freeze but not bustle/dbus-monitor.

A6GibKm:
I think shutdown will write the eof byte and we might be just waiting forever if it was not set. What i dont follow is that close should implicitly call shutdown.

src/desktop/secret.rs
```diff
@@ -93,11 +93,23 @@ impl<'a> std::ops::Deref for Secret<'a> {
...
+        let (x1, mut x2) = UnixStream::pair()?;
+        proxy.retrieve(&x2).await?;
+        x2.shutdown().await?;
+        x1
+    };
+    #[cfg(feature = "async-std")]
+    let mut x1 = {
+        let (x1, x2) = UnixStream::pair()?;
+        proxy.retrieve(&x2).await?;
+        x2.shutdown(Shutdown::Both)?;
```
A6GibKm, line 109: Afaik, there is no need to close both halves, closing the writing half should suffice.

A6GibKm:
@SeaDve @bilelmoussaoui Could you please verify that this actually fixes the issue?

The reproducer is to call the retrieve method while bustle 0.9 (note that flathub is at 0.10) is recording the session bus.

bilelmoussaoui:
> @SeaDve @bilelmoussaoui Could you please verify that this actually fixes the issue?
> 
> The reproducer is to call the retrieve method while bustle 0.9 (note that flathub is at 0.10) is recording the session bus.

I have it on my todo list. will try to have a look the upcoming days

A6GibKm:
Can confirm that this fixes the issues when tested with ASHPD demo (async-std) and Bustle on the commit before migrating to dbus-monitor and without this MR it does not work.

PR #238, opened 2024-09-22
Use the crate from the repo and async-std

hfiguiere:
1. so that we always build with the latest. 
2. fix feature so that it actually work.

### DIFF
--- a/ashpd-demo/Cargo.lock
+++ b/ashpd-demo/Cargo.lock
@@ -44,9 +44,7 @@ checksum = "b3d1d046238990b9cf5bcde22a3fb3584ee5cf65fb2765f454ed428c7a0063da"
 
 [[package]]
 name = "ashpd"
-version = "0.9.1"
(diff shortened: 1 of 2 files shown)


A6GibKm:
Unfortunately, it is not possible to build the demo with GNOME Builder due to the same error seen on CI.

bilelmoussaoui:
The demo is never meant to use the git version of ashpd. Because of two reasons:
- The one stated above by Maximilano. You can't make it work without opening a sandbox hole
- When a new breaking API release of gtk4-rs is around the corner, we switch to use it in ashpd main allowing people to try the latest APIs, if we ever merge something like this PR, I will have to update the demo continuously which something I don't want to.

I know it makes things easier for developing a new portal implementation, by having a very easy way to test it. But you don't really need a GUI for it.

The GUI is targeting distributions and users to debug portals, not writing ones.

Issue #155, opened 2023-08-17
Grabbing frames from the PipeWire stream isn't explained.


Frostie314159:
There is no documentation on how to actually read frames from the PipeWire stream acquired with screencast.

bilelmoussaoui:
Feel free to submit a patch. I am not a pipewire hacker, I only pass the fd & node_id to gstreamer :)

A6GibKm:
> ashpd-demo isn't very clear on this.

Is basically *the* example.

bilelmoussaoui:
Have you checked https://github.com/bilelmoussaoui/ashpd/blob/master/ashpd-demo/src/widgets/gst_paintable.rs? there is no "small" example as far as i know, feel free to build one :)

SyedAhkam:
Thanks for the prompt response. Yes. I had checked out `gst_paintable.rs` but as I said it wasn't too clear mixed with UI logic and believe me I tried my hardest to make this work the past few days, either with gstreamer or pipewire (but honestly more with pipewire)

Finally got some success right after I commented here - so basically I ended up mixing this example from [pipewire-rs](https://gitlab.freedesktop.org/pipewire/pipewire-rs/-/blob/86df39190c0ab67444666a42908f7e8c1344e24a/pipewire/examples/streams.rs) into the docs.rs example for ashpd on `Screencast` and came up with [this](https://gist.github.com/SyedAhkam/6131487ea9caaf6113426ecfb7104a75).

Although my attempt was non-gstreamer based, I think it solves the crux of this issue. If it looks good to you guys, I can go ahead and make a PR to include this as a standalone example?

bilelmoussaoui:
That would be great, please do so :)

Frostie314159:
I totally forgot about this issue. I'm guessing I can't help a lot with it, but it's nice to see this resolved!

PR #247, opened 2024-11-18
feat: add screen cast with pipewire example

SyedAhkam:
fixes: #155

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -64,6 +64,8 @@ zbus = { version = "5.0", default-features = false, features = ["url"] }
 [dev-dependencies]
 serde_json = "1.0"
 reis = { version = "0.4", features = ["tokio"] }
+pipewire = "0.8.0"
+tokio = { version = "1.41", features = [ "rt-multi-thread", "macros" ] }
 
 [package.metadata.docs.rs]
 features = ["backend", "gtk4", "raw_handle"]
--- a/README.md
+++ b/README.md
@@ -37,14 +37,6 @@ pub async fn run() -> ashpd::Result<()> {
 }
 ```
 
-Open a Screencast stream and connect using [pipewire-rs](https://gitlab.freedesktop.org/pipewire/pipewire-rs/)
-
-```bash
-$ cargo run --example screen_cast_pw
-```
-
-The source can be read [here](examples/screen_cast_pw.rs).
-
 ## Optional features
 
 | Feature | Description | Default |
--- a/README.md
+++ b/README.md
@@ -37,6 +37,14 @@ pub async fn run() -> ashpd::Result<()> {
 }
 ```
 
+Open a Screencast stream and connect using [pipewire-rs](https://gitlab.freedesktop.org/pipewire/pipewire-rs/)
+
+```bash
+$ cargo run --example screen_cast_pw
+```
+
+The source can be read [here](examples/screen_cast_pw.rs).
+
 ## Optional features
 
 | Feature | Description | Default |
--- a/examples/screen_cast_pw.rs
+++ b/examples/screen_cast_pw.rs
@@ -1,0 +1,218 @@
+use std::os::fd::{IntoRawFd, OwnedFd};
+
+use ashpd::desktop::{
+    screencast::{CursorMode, Screencast, SourceType, Stream as AshStream},
+    PersistMode,
+};
+use pipewire as pw;
+use pw::{properties::properties, spa};
+
+use spa::pod::Pod;
+
+struct UserData {
+    format: spa::param::video::VideoInfoRaw,
+}
+
+async fn open_portal() -> ashpd::Result<(AshStream, OwnedFd)> {
+    let proxy = Screencast::new().await?;
+    let session = proxy.create_session().await?;
+    proxy
+        .select_sources(
+            &session,
+            CursorMode::Hidden,
+            SourceType::Monitor.into(),
+            false,
+            None,
+            PersistMode::DoNot,
+        )
+        .await?;
+
+    let response = proxy.start(&session, None).await?.response()?;
+    let stream = response.streams().first().unwrap().to_owned();
+
+    let fd = proxy.open_pipe_wire_remote(&session).await?;
+
+    Ok((stream, fd))
+}
+
+async fn start_streaming(node_id: u32) -> Result<(), pw::Error> {
+    pw::init();
+
+    let mainloop = pw::main_loop::MainLoop::new(None)?;
+    let context = pw::context::Context::new(&mainloop)?;
+    let core = context.connect(None)?;
+
+    let data = UserData {
+        format: Default::default(),
+    };
+
+    let stream = pw::stream::Stream::new(
+        &core,
+        "video-test",
+        properties! {
+            *pw::keys::MEDIA_TYPE => "Video",
+            *pw::keys::MEDIA_CATEGORY => "Capture",
+            *pw::keys::MEDIA_ROLE => "Screen",
+        },
+    )?;
+
+    let _listener = stream
+        .add_local_listener_with_user_data(data)
+        .state_changed(|_, _, old, new| {
+            println!("State changed: {:?} -> {:?}", old, new);
+        })
+        .param_changed(|_, user_data, id, param| {
+            let Some(param) = param else {
+                return;
+            };
+            if id != pw::spa::param::ParamType::Format.as_raw() {
+                return;
+            }
+
+            let (media_type, media_subtype) =
+                match pw::spa::param::format_utils::parse_format(param) {
+                    Ok(v) => v,
+                    Err(_) => return,
+                };
+
+            if media_type != pw::spa::param::format::MediaType::Video
+                || media_subtype != pw::spa::param::format::MediaSubtype::Raw
+            {
+                return;
+            }
+
+            user_data
+                .format
+                .parse(param)
+                .expect("Failed to parse param changed to VideoInfoRaw");
+
+            println!("got video format:");
+            println!(
+                "  format: {} ({:?})",
+                user_data.format.format().as_raw(),
+                user_data.format.format()
+            );
+            println!(
+                "  size: {}x{}",
+                user_data.format.size().width,
+                user_data.format.size().height
+            );
+            println!(
+                "  framerate: {}/{}",
+                user_data.format.framerate().num,
+                user_data.format.framerate().denom
+            );
+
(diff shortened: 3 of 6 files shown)


examples/screen_cast_pw.rs
```diff
@@ -0,0 +1,218 @@
+use std::os::fd::{IntoRawFd, OwnedFd};
+
+use ashpd::desktop::{
+    screencast::{CursorMode, Screencast, SourceType, Stream as AshStream},
```
bilelmoussaoui, line 4:
```suggestion
    screencast::{CursorMode, Screencast, SourceType, Stream as ScreencastStream},
```
instead?

examples/screen_cast_pw.rs
```diff
@@ -0,0 +1,218 @@
...
+    PersistMode,
+};
+use pipewire as pw;
```
bilelmoussaoui, line 7: Maybe rename the crate in Cargo.toml, using `pw = {package = "pipewire"...}`?

examples/screen_cast_pw.rs
```diff
@@ -0,0 +1,218 @@
...
+use pw::{properties::properties, spa};
+
+use spa::pod::Pod;
```
bilelmoussaoui, line 10:
```suggestion
use pw::{properties::properties, spa::{self, pod::Pod}};
```

examples/screen_cast_pw.rs
```diff
@@ -0,0 +1,218 @@
...
+            CursorMode::Hidden,
+            SourceType::Monitor.into(),
+            false,
+            None,
+            PersistMode::DoNot,
+        )
+        .await?;
+
+    let response = proxy.start(&session, None).await?.response()?;
+    let stream = response.streams().first().unwrap().to_owned();
```
bilelmoussaoui, line 31: Should maybe check first if there are any streams? but as it is an example, maybe that is fine

examples/screen_cast_pw.rs
```diff
@@ -0,0 +1,218 @@
...
+            println!(
+                "  size: {}x{}",
+                user_data.format.size().width,
+                user_data.format.size().height
+            );
+            println!(
+                "  framerate: {}/{}",
+                user_data.format.framerate().num,
+                user_data.format.framerate().denom
+            );
```
bilelmoussaoui, line 104:
```suggestion
            println!("got video format");
            println!(
                "\tformat: {} ({:?})",
                user_data.format.format().as_raw(),
                user_data.format.format()
            );
            println!(
                "\tsize: {}x{}",
                user_data.format.size().width,
                user_data.format.size().height
            );
            println!(
                "\tframerate: {}/{}",
                user_data.format.framerate().num,
                user_data.format.framerate().denom
            );
```

Maybe?

examples/screen_cast_pw.rs
SyedAhkam: well we could but our users might get confused the moment they try to copy & paste the source directly in their project? their setup will by default not resolve it correctly?

examples/screen_cast_pw.rs
SyedAhkam: just getting rid of this distinction since Pod only had a single reference  anyway; qualifying it as spa::pod::Pod now.

examples/screen_cast_pw.rs
SyedAhkam: I guess I can turn the unwrap into an expect with an error message

examples/screen_cast_pw.rs
SyedAhkam: yeah sure.

examples/screen_cast_pw.rs
bilelmoussaoui: well, that is true :) but it is something i expect a rust user to be familiar with imho

examples/screen_cast_pw.rs
bilelmoussaoui: that would be better indeed

examples/screen_cast_pw.rs
SyedAhkam: haha alright sure

SyedAhkam: looks like the `pw::spa::pod::object!` macro from pipewire expects the module `pipewire` to exist

```rust
#[doc(hidden)]
#[macro_export]
macro_rules! __object__ {
    ($type_:expr, $id:expr, $($properties:expr),* $(,)?) => {
        pipewire::spa::pod::Object { // <-- here
            type_: $type_.as_raw(),
            id: $id.as_raw(),
            properties: [ $( $properties, )* ].to_vec(),
        }
    };
}
```

and.. a fix is to do the opposite 🙃

```
use pw as pipewire
```

they should've used $crate instead lol

----------------------------

now should we umm report this to pipewire folks or just leave pipewire as is?

bilelmoussaoui: it would be nice to report it nonetheless but for now i think just using pipewire as pw is ok :)

SyedAhkam: cool! I'll report it on the side

README.md
```diff
@@ -37,6 +37,14 @@ pub async fn run() -> ashpd::Result<()> {
 }
 ```
 
+Open a Screencast stream and connect using [pipewire-rs](https://gitlab.freedesktop.org/pipewire/pipewire-rs/)
+
+```bash
+$ cargo run --example screen_cast_pw
+```
+
+The source can be read [here](examples/screen_cast_pw.rs).
```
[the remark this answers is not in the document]
SyedAhkam: image
image

thoughts?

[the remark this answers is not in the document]
bilelmoussaoui: yep, looks good to me!

SyedAhkam:
All done, lets wait for your friend now.

examples/screen_cast_pw.rs
```diff
@@ -0,0 +1,220 @@
...
+    println!("Connected stream");
+
+    mainloop.run();
+
+    Ok(())
+}
+
+#[tokio::main]
+async fn main() {
+    let (stream, fd) = open_portal().await.expect("failed to open portal");
```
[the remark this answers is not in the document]
bilelmoussaoui: I think it would cause an issue if you have an app starting a screencast/camera stream at once. But i don't know enough about pipewire to say for sure. I would say, use connect_fd

examples/screen_cast_pw.rs
```diff
@@ -0,0 +1,218 @@
...
+            Choice,
+            Enum,
+            Id,
+            pw::spa::param::video::VideoFormat::RGB,
+            pw::spa::param::video::VideoFormat::RGB,
+            pw::spa::param::video::VideoFormat::RGBA,
+            pw::spa::param::video::VideoFormat::RGBx,
+            pw::spa::param::video::VideoFormat::BGRx,
+            pw::spa::param::video::VideoFormat::YUY2,
+            pw::spa::param::video::VideoFormat::I420,
```
rmader, line 153: I'd suggest to have all 8 32-bit rgb formats (https://gitlab.freedesktop.org/pipewire/pipewire/-/blob/master/spa/include/spa/param/video/raw.h?ref_type=heads#L45-52) in the list here so compositors/drivers can choose the optimal one.

Also the duplicated `RGB` at the start means that it`s the preferred format - which is not great as 24-bit RGB can be really slow on modern GPUs. I'd suggest to drop `RGB` altogether - or at least but some 32-bit format at the top.

rmader (commented):
Had a look over the PW parts - only one comment, the rest look sensible to me

examples/screen_cast_pw.rs
SyedAhkam: Thanks for your input, this is quite insightful! @rmader 

How about this sequence then? I've de-prioritised RGB based formats to death.

```
pw::spa::param::video::VideoFormat::BGRx,
pw::spa::param::video::VideoFormat::BGRx,
pw::spa::param::video::VideoFormat::xBGR,
pw::spa::param::video::VideoFormat::BGRA,
pw::spa::param::video::VideoFormat::ABGR,
pw::spa::param::video::VideoFormat::YUY2,
pw::spa::param::video::VideoFormat::I420,
pw::spa::param::video::VideoFormat::RGBA,
pw::spa::param::video::VideoFormat::ARGB,
pw::spa::param::video::VideoFormat::xRGB,
pw::spa::param::video::VideoFormat::RGBx,
pw::spa::param::video::VideoFormat::RGBA,
pw::spa::param::video::VideoFormat::RGBx,
pw::spa::param::video::VideoFormat::RGB,
```

rmader: Only the first format should matter - that's the default pick (it has to be listed again as option afterwards). The rest is up to you, but it might make sense to follow some order, such as:

```
pw::spa::param::video::VideoFormat::BGRx,
pw::spa::param::video::VideoFormat::ABGR,
pw::spa::param::video::VideoFormat::xBGR,
pw::spa::param::video::VideoFormat::ARGB,
pw::spa::param::video::VideoFormat::xRGB,
pw::spa::param::video::VideoFormat::BGRA,
pw::spa::param::video::VideoFormat::BGRx,
pw::spa::param::video::VideoFormat::RGBA,
pw::spa::param::video::VideoFormat::RGBx,
pw::spa::param::video::VideoFormat::RGB,
pw::spa::param::video::VideoFormat::I420,
pw::spa::param::video::VideoFormat::YUY2,
```

SyedAhkam: we can keep BGRx as the default pick, yeah?

Cargo.toml
```diff
@@ -64,6 +64,8 @@ zbus = { version = "5.0", default-features = false, features = ["url"] }
 [dev-dependencies]
 serde_json = "1.0"
 reis = { version = "0.4", features = ["tokio"] }
+pipewire = "0.8.0"
+tokio = { version = "1.41", features = [ "rt-multi-thread", "macros" ] }
```
hfiguiere, line 68: can't we use async-std instead? It's already a dependency.

SyedAhkam: Yeah, even in my original example- I had used async_std but I saw this crate is mostly based on tokio so tokio made more sense. Plus many users will majorly be based on tokio only improving the copy/pasting ability.

examples/screen_cast_pw.rs
bilelmoussaoui: Let us land this as is, if default format could be improved, it can be done in a separate PR.

PR #249, opened 2024-11-26
demo: Allow to set file chooser filters

agx:
Add a combox box that allows select different filters to the OpenFile and SaveFile dialogs. Adding different mime types allows for some flexibility when testing e.g. file choser implementations.
    
When a filter set of filters is selected we pick the first filter as the `current_filter`.
The default is no filters.

A separate commit avoids a redirect when cloning libshumate

Screenshot:

filters

### DIFF
--- a/ashpd-demo/src/portals/desktop/file_chooser.rs
+++ b/ashpd-demo/src/portals/desktop/file_chooser.rs
@@ -1,6 +1,6 @@
 use adw::{prelude::*, subclass::prelude::*};
 use ashpd::{
-    desktop::file_chooser::{OpenFileRequest, SaveFileRequest, SaveFilesRequest},
+    desktop::file_chooser::{OpenFileRequest, SaveFileRequest, SaveFilesRequest, FileFilter},
     WindowIdentifier,
 };
 use gtk::glib;
@@ -27,6 +27,8 @@ mod imp {
         #[template_child]
         pub open_directory_switch: TemplateChild<adw::SwitchRow>,
         #[template_child]
+        pub open_filter_combo: TemplateChild<adw::ComboRow>,
+        #[template_child]
         pub open_response_group: TemplateChild<adw::PreferencesGroup>,
 
         #[template_child]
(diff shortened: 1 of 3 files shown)


ashpd-demo/src/portals/desktop/file_chooser.rs
```diff
@@ -107,13 +127,19 @@ impl FileChooserPage {
         let modal = imp.open_modal_switch.is_active();
         let multiple = imp.open_multiple_switch.is_active();
         let accept_label = is_empty(imp.open_accept_label_entry.text());
-
+        let filters = self.filters(imp.open_filter_combo.selected());
+        let current_filter = match filters.len() {
```
A6GibKm, line 131: let current_filter = filters.first()

bilelmoussaoui:
This only supports specific mimetypes without allowing testing the glob pattern support. Would you mind showing a screenshot how it looks?

agx:
> This only supports specific mimetypes without allowing testing the glob pattern support. 

In fact I had a `glob("*.txt")` in there before. I can add that back if it's useful.

> Would you mind showing a screenshot how it looks?

Added a screenshot to the MR description.

agx:
> In fact I had a glob("*.txt") in there before. I can add that back if it's useful.

Added a `.glob("*.txt")` to test the glob path as well

ashpd-demo/src/portals/desktop/file_chooser.rs
```diff
@@ -107,13 +127,16 @@ impl FileChooserPage {
         let modal = imp.open_modal_switch.is_active();
         let multiple = imp.open_multiple_switch.is_active();
         let accept_label = is_empty(imp.open_accept_label_entry.text());
-
```
bilelmoussaoui, line 110: unneeded change

bilelmoussaoui (approved):
otherwise lgtm!

bilelmoussaoui:
> Added a screenshot to the MR description.

I don't see one but that is okay i guess :)

bilelmoussaoui:
Thanks!

PR #252, opened 2024-12-18
backend: Pass handle token to method calls for easier identification and association with their backend logic

arun-mani-j:
In the current state of `ashpd::backend`, it is difficult to associate each request with the backend logic. So any further calls on the request is left in ambiguity.

For example, let's say two apps call `Account.GetUserInformation`. Our implementation shows dialogs that lets the user enter their details. Now, if one of these apps cancel their request with `Request.close()`, then it is impossible (or difficult) for implementer to know which dialog must be closed.

An easy fix to this problem is to pass the object path of request on the method calls. Now the backend can associate each request with its backend logic, like say a hash map of requests to dialogs. Then, when `close` method is called, it can identify which dialog to close and can close it.

I'm a Rust newbie, so not sure if `clone()`ing paths are a good idea or if we can tackle this problem from a different angle. :sweat_smile: 
Thank you

### DIFF
--- a/backend-demo/src/settings.rs
+++ b/backend-demo/src/settings.rs
@@ -5,7 +5,10 @@ use ashpd::{
         request::RequestImpl,
         settings::{SettingsImpl, SettingsSignalEmitter},
     },
-    desktop::settings::{ColorScheme, Namespace, APPEARANCE_NAMESPACE, COLOR_SCHEME_KEY},
+    desktop::{
+        settings::{ColorScheme, Namespace, APPEARANCE_NAMESPACE, COLOR_SCHEME_KEY},
+        HandleToken,
+    },
     zbus::zvariant::OwnedValue,
     PortalError,
 };
@@ -20,8 +23,8 @@ pub struct Settings {
 
 #[async_trait]
 impl RequestImpl for Settings {
-    async fn close(&self) {
-        tracing::debug!("IN Close()");
+    async fn close(&self, token: HandleToken) {
+        tracing::debug!("IN Close(): {token}");
     }
 }
 
--- a/src/backend/access.rs
+++ b/src/backend/access.rs
@@ -7,7 +7,7 @@ use crate::{
         request::{Request, RequestImpl},
         MaybeAppID, MaybeWindowIdentifier, Result,
     },
-    desktop::{file_chooser::Choice, request::Response, Icon},
+    desktop::{file_chooser::Choice, request::Response, HandleToken, Icon},
     zvariant::{self, DeserializeDict, OwnedObjectPath, SerializeDict},
     AppID, WindowIdentifierType,
 };
@@ -63,8 +63,10 @@ impl AccessResponse {
 
 #[async_trait]
 pub trait AccessImpl: RequestImpl {
+    #[allow(clippy::too_many_arguments)]
     async fn access_dialog(
         &self,
+        token: HandleToken,
         app_id: Option<AppID>,
         window_identifier: Option<WindowIdentifierType>,
         title: String,
@@ -109,10 +111,11 @@ impl AccessInterface {
         Request::spawn(
             "Access::AccessDialog",
             &self.cnx,
-            handle,
+            handle.clone(),
             Arc::clone(&self.imp),
             async move {
                 imp.access_dialog(
+                    HandleToken::try_from(&handle).unwrap(),
                     app_id.inner(),
                     window_identifier.inner(),
                     title,
--- a/src/backend/print.rs
+++ b/src/backend/print.rs
@@ -10,6 +10,7 @@ use crate::{
     desktop::{
         print::{PageSetup, PreparePrint, Settings},
         request::Response,
+        HandleToken,
     },
     zvariant::{self, DeserializeDict, OwnedObjectPath},
     AppID, WindowIdentifierType,
@@ -51,8 +52,10 @@ impl PrintOptions {
 
 #[async_trait]
 pub trait PrintImpl: RequestImpl {
+    #[allow(clippy::too_many_arguments)]
     async fn prepare_print(
         &self,
+        token: HandleToken,
         app_id: Option<AppID>,
         parent_window: Option<WindowIdentifierType>,
         title: String,
@@ -63,6 +66,7 @@ pub trait PrintImpl: RequestImpl {
 
     async fn print(
         &self,
+        token: HandleToken,
         app_id: Option<AppID>,
         parent_window: Option<WindowIdentifierType>,
         title: String,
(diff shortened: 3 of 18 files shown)


src/backend/request.rs
```diff
@@ -39,9 +39,10 @@ impl Request {
         #[cfg(feature = "tracing")]
         tracing::debug!("{_method}");
         let (fut, abort_handle) = abortable(callback);
+        let path_clone = path.clone();
```
bilelmoussaoui, line 42: I don't really think the path must be passed to every function there is, i am kinda wondering if we should instead pass the AbortHandle...but as it doesn't implement hash, you would not be able to construct a mapping between spawned dialogs and the corresponding requests.

arun-mani-j: My reasoning for path is that - it is really simple solution this way and we can get the path from incoming D-Bus request itself without having to store any state in the library side.

Also whatever we are passing for the interface request (like `Account.GetUserInformation`) we need to pass to `Request.close` too so as to establish the association.

Happy to try better ideas and workarounds :)

bilelmoussaoui: The thing is I really don't want to expose the Request object path, if we really have to expose part of it, I would expose a HandleToken which we can retrieve from the last portion of the ObjectPath. Basically `HandleToken::try_from(path.as_str().split('/').last().unwrap()).unwrap()` or so.

Where are you using the backend feature btw?

arun-mani-j: I think the `HandleToken`s also suit here. Do you have any specific suggestions on how we can expose it to the impls? May be we add a new constructor to `HandleToken` which lets us construct from an `ObjectPath` or so?

We are trying to rewrite Phosh portal in Rust and that's where we are using `ashpd` :).

bilelmoussaoui: Interesting! let me know how the experiment goes. 

I would do a TryFrom<ObjectPath> for HandleToken that is gated by the backend feature.

arun-mani-j:
v1:
- Change from using object path to handle token.

src/desktop/mod.rs
```diff
@@ -1,8 +1,8 @@
 mod handle_token;
 pub(crate) mod request;
 mod session;
-pub(crate) use self::handle_token::HandleToken;
 pub use self::{
+    handle_token::HandleToken,
```
bilelmoussaoui, line 5: Should be only exposed if backend is enabled

bilelmoussaoui:
HandleToken also needs to derive various traits like Hash, PartialEq,Eq and so on

backend-demo/src/settings.rs
```diff
@@ -5,7 +5,10 @@ use ashpd::{
         request::RequestImpl,
         settings::{SettingsImpl, SettingsSignalEmitter},
     },
-    desktop::{settings::{ColorScheme, Namespace, APPEARANCE_NAMESPACE, COLOR_SCHEME_KEY}, HandleToken},
+    desktop::{
```
bilelmoussaoui, line 8: the formatting fixes should be squashed with the commits that caused them.

src/desktop/handle_token.rs
```diff
@@ -84,6 +87,16 @@ impl TryFrom<&str> for HandleToken {
     }
 }
 
+#[cfg(feature = "backend")]
+impl TryFrom<&OwnedObjectPath> for HandleToken {
+    type Error = HandleInvalidCharacter;
+
+    fn try_from(value: &OwnedObjectPath) -> Result<Self, Self::Error> {
+        let base_segment = value.as_str().split('/').last();
+        HandleToken::try_from(base_segment.unwrap())
```
bilelmoussaoui, line 96:
```suggestion
        let base_segment = value.as_str().split('/').last().expect("A valid request ObjectPath");
        HandleToken::try_from(base_segment)
```

src/backend/print.rs
```diff
@@ -53,6 +54,7 @@ impl PrintOptions {
 pub trait PrintImpl: RequestImpl {
     async fn prepare_print(
```
bilelmoussaoui, line 55: Silence the clippy warning

bilelmoussaoui (commented):
otherwise lgtm

src/desktop/handle_token.rs
```diff
@@ -5,7 +5,10 @@ use std::{
 
 use rand::{distributions::Alphanumeric, thread_rng, Rng};
 use serde::{Deserialize, Serialize};
-use zbus::{names::OwnedMemberName, zvariant::Type};
+use zbus::{
+    names::OwnedMemberName,
+    zvariant::{OwnedObjectPath, Type},
```
bilelmoussaoui, line 10: This is only usef if the backend feature is enabled, causing a conditional clippy warning

arun-mani-j:
v2:
- HandleToken:
  - Derive Eq, PartialEq and Hash on `backend` feature.
  - Use `.expect` while converting from path.
  - Fix clippy warning on used imports on non-`backend`.
- Desktop:
  - Export `HandleToken` only on `backend` feature.
- Access and Print of Backend:
  - Silence Clippy warning of too many arguments.
- Tree:
  - Perform indent in the same commit.

src/backend/access.rs
```diff
@@ -61,10 +61,12 @@ impl AccessResponse {
     }
 }
 
+#[allow(clippy::too_many_arguments)]
 #[async_trait]
 pub trait AccessImpl: RequestImpl {
     async fn access_dialog(
```
bilelmoussaoui, line 67:
```suggestion
#[async_trait]
pub trait AccessImpl: RequestImpl {
    #[allow(clippy::too_many_arguments)]
    async fn access_dialog(
```
instead

src/backend/print.rs
```diff
@@ -49,10 +50,12 @@ impl PrintOptions {
     }
 }
 
+#[allow(clippy::too_many_arguments)]
 #[async_trait]
 pub trait PrintImpl: RequestImpl {
     async fn prepare_print(
```
bilelmoussaoui, line 56:
```suggestion
#[async_trait]
pub trait PrintImpl: RequestImpl {
    #[allow(clippy::too_many_arguments)]
    async fn prepare_print(
```
and here

arun-mani-j:
v3:
- Access and Print of backend:
  - Silence clippy only for the method.

src/backend/access.rs
arun-mani-j: That's where I wanted to put too, but by mistake put it over the trait. Thanks.

bilelmoussaoui:
Thanks!
Won't release this anytime soon as it contains breaking API changes. Feel free to use ashpd from git and find what other missing/broken things in ashpd that would make implementing the posh portals easier.

arun-mani-j:
Thank you!
We are iterating the plan for Phosh's new portal in Rust at https://gitlab.gnome.org/arun-mani-j/xdg-desktop-portal-phosh/-/merge_requests/1. 
Thanks again for the support and ashpd :)

PR #256, opened 2025-01-07
feat: error when neither tokio nor async-std are enabled

FineFindus:
Shows a compiler warning when neither tokio nor async-std are enabled. When both features are disabled, the project fails to compile, as the symbols cannot be resolved.


bilelmoussaoui:
Thanks

PR #257, opened 2025-01-14
 Add API for setting host apps AppID 

bilelmoussaoui:


GeorgesStavracas (approved):

src/desktop/camera.rs
```diff
@@ -17,8 +17,8 @@
 //! }
 //! ```
 //! An example on how to connect with Pipewire can be found [here](https://github.com/bilelmoussaoui/ashpd/blob/master/examples/screen_cast_pw.rs).
-//! Although the example's primary focus is screen casting, stream connection logic remains the same -- with one accessibility change:
-//! ```rust,ignore
+//! Although the example's primary focus is screen casting, stream connection
+//! logic remains the same -- with one accessibility change: ```rust,ignore
```
A6GibKm, line 21: The triple backtick being in the same line looks wrong to me.

src/registry.rs
```diff
@@ -0,0 +1,47 @@
...
+impl<'a> RegistryProxy<'a> {
+    pub async fn new() -> Result<RegistryProxy<'a>, Error> {
+        let proxy = Proxy::new_desktop("org.freedesktop.host.portal.Registry").await?;
+        Ok(Self(proxy))
+    }
+
+    pub async fn register(&self, app_id: AppID) -> Result<(), Error> {
+        let options = RegisterOptions {};
+        self.0.call_method("Register", &(&app_id, &options)).await?;
+        return Ok(());
```
A6GibKm, line 20: Just plain `Ok(())` without a `return` will do here.

src/registry.rs
```diff
@@ -0,0 +1,47 @@
...
+}
+/// Registers a host application for portal usage.
+///
+/// Portals rely on the application ID to store and manage the permissions of
+/// individual apps. However, for non-sandboxed (host) applications, this
+/// information cannot be directly retrieved. To resolve this, the method first
+/// verifies that the application is not sandboxed. It then uses the
+/// `org.freedesktop.host.portal.Registry` interface to register the process
+/// communicating over DBus with the portal as the owner of the specified
+/// application ID. For more technical details, see ...
```
A6GibKm, line 39: On thing I am missing is when to call this, e.g. we had problems with the camera permission if it was called too soon. cc @GeorgesStavracas

A6GibKm: Also the `...` should be replaced before merging.

A6GibKm:
For anyone wondering, this depends on https://github.com/flatpak/xdg-desktop-portal/pull/1521

checkraisefold:
xdg-desktop-portal PR merged

PR #263, opened 2025-01-19
desktop/proxy: Properly propagate PortalError

### LINKED ISSUE #261: request_access does not produce PortalErrors
The fn `request_access` in the camera portal uses `request_access` and instead of returning a `PortalError::NotAllowed`, it returns a `ashpd::Error::ZBus(...)` containing `org.freedesktop.portal.Error.NotAllowed`.

bilelmoussaoui:
Fixes #261


A6GibKm:
I can confirm this fixes the issue. However it begs the question of where else are we missing this.

bilelmoussaoui:
That is the only possibile place i believe

PR #266, opened 2025-02-14
Restructure project

bilelmoussaoui:
By making the demo clearly separate from the project


bilelmoussaoui:
This means making tokio default for the client demo, annoying. Dropping the idea for now

PR #270, opened 2025-02-27
Fix links to source code in comments and documentation

oSoMoN:
Make them valid permalinks that will always point to the intended piece of code.

### DIFF
--- a/src/desktop/camera.rs
+++ b/src/desktop/camera.rs
@@ -89,7 +89,7 @@ impl<'a> Camera<'a> {
     #[doc(alias = "xdp_portal_open_pipewire_remote_for_camera")]
     pub async fn open_pipe_wire_remote(&self) -> Result<OwnedFd, Error> {
         // `options` parameter doesn't seems to be used yet
-        // see https://github.com/flatpak/xdg-desktop-portal/blob/master/src/camera.c#L178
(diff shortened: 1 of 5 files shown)


A6GibKm:
Please use a tag with the latest released stable version instead of a commit.

oSoMoN:
> Please use a tag with the latest released stable version instead of a commit.

Done.

bilelmoussaoui:
There is 0 chance to keep this up to date as the links will always be outdated (code moves around) and it is meant to give you an indication that things might have changed and needs an update from our side. Just pointing the links to something that is always valid is not the intent here...

oSoMoN:
I'm not following you. In master right now, the links are not pointing to where they were pointing when you wrote the code originally, so they are indeed outdated, and this may confuse readers of these comments.

The change I'm proposing makes them point to the relevant code at a given point in time, so they will remain valid. Note that I updated the line numbers to make sure they're actually pointing to where they were pointing when you wrote that code, and I made sure the code being referenced matches the comments.

There is no need to continuously update those links with newer releases of xdg-desktop-portal and gtk, if that's what you meant by "to keep this up to date".

bilelmoussaoui:
I am still not very sure about this, but let us merge it if it would make you happy.

oSoMoN:
> I am still not very sure about this, but let us merge it if it would make you happy.

Thanks. If there's anything I can do to update this change (in a followup PR) to make it more convincing to you, let me know.

PR #271, opened 2025-03-07
backend-demo: Add documentation

oSoMoN:

PR #274, opened 2025-03-10
backend-demo: Implement the Access interface

oSoMoN:
Without it, the xdg-desktop-portal frontend will skip looking for an
implementation of the Screenshot and Wallpaper interfaces, among others
(see https://github.com/flatpak/xdg-desktop-portal/blob/2fb76ffb/src/xdg-desktop-portal.c#L321-L358).

### DIFF
--- a/backend-demo/src/access.rs
+++ b/backend-demo/src/access.rs
@@ -1,0 +1,40 @@
+use ashpd::{
+    backend::{
+        access::{AccessImpl, AccessOptions, AccessResponse},
+        request::RequestImpl,
+        Result,
+    },
+    desktop::HandleToken,
+    AppID, WindowIdentifierType,
+};
+use async_trait::async_trait;
+
+#[derive(Default)]
+pub struct Access;
+
+#[async_trait]
+impl RequestImpl for Access {
+    async fn close(&self, token: HandleToken) {
+        tracing::debug!("IN Close(): {token}");
+    }
+}
+
+#[async_trait]
+impl AccessImpl for Access {
+    async fn access_dialog(
+        &self,
+        _handle: HandleToken,
+        _app_id: Option<AppID>,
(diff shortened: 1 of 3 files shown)


oSoMoN:
Note: this is stacked on top of #271 as the issue is made more visible when following the instructions in that other PR, but there is no strict dependency between the two.

bilelmoussaoui:
needs a rebase

oSoMoN:
> needs a rebase

Done.

backend-demo/src/access.rs
```diff
@@ -0,0 +1,40 @@
...
+        &self,
+        _handle: HandleToken,
+        _app_id: Option<AppID>,
+        _window_identifier: Option<WindowIdentifierType>,
+        _title: String,
+        _subtitle: String,
+        _body: String,
+        options: AccessOptions,
+    ) -> Result<AccessResponse> {
+        let mut response = <AccessResponse as std::default::Default>::default();
```
bilelmoussaoui, line 34:
```suggestion
        let mut response = AccessResponse::default();
```

backend-demo/src/access.rs
```diff
@@ -0,0 +1,40 @@
...
+        for choice in options.choices() {
+            response = response.choice(choice.id(), choice.initial_selection());
+        }
```
bilelmoussaoui, line 37: Do we really want to do this?

backend-demo/src/access.rs
oSoMoN: Applied.

backend-demo/src/access.rs
oSoMoN: That's admittedly a dummy implementation that responds with the default (initial) selection. What other behaviour that doesn't involve user interaction would you suggest?

bilelmoussaoui:
Thanks!

PR #285, opened 2025-04-14
backend: Always release a request even when it returns an error

### LINKED ISSUE #284: backend request isn't released if it returns an error
If a portal backend returns a `PortalError`, the request object is never released, it remains dangling and exported on the bus.

This is caused by [the use of the ? operator](https://github.com/bilelmoussaoui/ashpd/blob/31595d452f408f31ddfeaa49ce03222edcbcf21f/src/backend/request.rs#L59) which propagates an error up the call stack before the `spawn` function has a chance to clean up and release the request.

oSoMoN:
Closes #284

### DIFF
--- a/src/backend/request.rs
+++ b/src/backend/request.rs
@@ -54,16 +54,18 @@ impl Request {
         );
         server.at(&path, request).await?;
 
-        let response = match fut.await {
-            Err(_) => Response::cancelled(),
-            Ok(response) => Response::ok(response?),
-        };
+        let response = fut.await;
+
         #[cfg(feature = "tracing")]
         tracing::debug!("{_method} returned {:#?}", response);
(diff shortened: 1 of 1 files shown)


src/backend/request.rs
```diff
@@ -54,16 +54,18 @@ impl Request {
...
         server.at(&path, request).await?;
 
-        let response = match fut.await {
-            Err(_) => Response::cancelled(),
-            Ok(response) => Response::ok(response?),
-        };
+        let response = fut.await;
+
         #[cfg(feature = "tracing")]
         tracing::debug!("{_method} returned {:#?}", response);
```
A6GibKm, line 60: This change effectively changes how this call logs. Did you test whether more or less the output is logged?

Issue #38, opened 2022-01-10
Provide server side of the portals interfaces


bilelmoussaoui:
This would allow people to write a portals implementations in rust.

Very low priority, just in case might want to work on this

PR #286, opened 2025-04-16
backend: Add Session and ScreenCast iface implementations (related to #38)

oSoMoN:
Also add a dummy ScreenCast backend implementation to backend-demo.


bilelmoussaoui:
awesome, thank you!

src/backend/session.rs
```diff
@@ -0,0 +1,62 @@
...
+}
+
+#[zbus::interface(
+    name = "org.freedesktop.impl.portal.Session",
+    proxy(gen_blocking = false)
+)]
+impl Session {
+    #[zbus(property(emits_changed_signal = "const"), name = "version")]
+    fn version(&self) -> u32 {
+        0
```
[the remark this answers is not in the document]
oSoMoN: The [interface specification](https://github.com/flatpak/xdg-desktop-portal/blob/15ad0c3c12038eb2d00fc2f84410bf85f04ee78b/data/org.freedesktop.impl.portal.Session.xml#L53) doesn't define a default value, and I couldn't find any call sites for the generated `xdp_impl_session_set_version()`.

When I tested the gnome backend, the `version` property of sessions it creates is 0. Hence this default value here. But I can bump it to 1 if it makes sense.

src/backend/session.rs
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    #[zbus(signal)]
+    async fn closed(signal_emitter: &SignalEmitter<'_>) -> zbus::Result<()>;
+}
+
+#[derive(SerializeDict, Type, Debug)]
+#[zvariant(signature = "dict")]
+pub struct CreateSessionResponse {
+    session_id: String,
```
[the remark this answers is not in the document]
oSoMoN: The specification isn't overly clear (except for the fact that [`session_id` is a string](https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.impl.portal.ScreenCast.html#org-freedesktop-impl-portal-screencast-createsession)).

Looking at the implementation, the session id appears to be defined as [the full path](https://github.com/flatpak/xdg-desktop-portal/blob/15ad0c3c12038eb2d00fc2f84410bf85f04ee78b/src/xdp-session.c#L364-L383). But the gnome backend's implementation of `org.freedesktop.impl.portal.ScreenCast.CreateSession` returns [an empty `results` vardict](https://gitlab.gnome.org/GNOME/xdg-desktop-portal-gnome/-/blob/a9ddd98b9bac2cc7e4651b74884d5d8ffce94f2b/src/screencast.c#L838-L842). So I guess this is up for interpretation, as there isn't an existing implementation to draw inspiration from.

src/backend/screencast.rs
```diff
@@ -0,0 +1,241 @@
...
+#[async_trait]
+pub trait ScreencastImpl: RequestImpl {
+    fn available_source_types(&self) -> BitFlags<SourceType>;
+
+    fn available_cursor_mode(&self) -> BitFlags<CursorMode>;
+
+    async fn create_session(
+        &self,
+        handle: OwnedObjectPath,
+        session_handle: OwnedObjectPath,
```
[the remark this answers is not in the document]
oSoMoN: In the backend API, handle and session handle are [full-fledged object paths](https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.impl.portal.ScreenCast.html#org-freedesktop-impl-portal-screencast-createsession) (constructed by the frontend from the `SENDER` and `TOKEN` parameters).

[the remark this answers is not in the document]
bilelmoussaoui: I am aware. Although, the only unique part of the path is the HandleToken, see what we expose in RequestImpl.

So we should do the same here and avoid dbus specifities if not necessary

src/backend/screencast.rs
```diff
@@ -0,0 +1,241 @@
...
+    }
+}
+
+#[zbus::interface(name = "org.freedesktop.impl.portal.ScreenCast")]
+impl ScreencastInterface {
+    #[zbus(
+        property(emits_changed_signal = "const"),
+        name = "AvailableSourceTypes"
+    )]
+    fn available_source_types(&self) -> u32 {
```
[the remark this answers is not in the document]
oSoMoN: For the interface macro to generate the relevant code, it would require the `From<BitFlags<SourceType, u32>>` trait to be implemented. `BitFlags` does implement [`TryFrom<u32>`](https://docs.rs/enumflags2/latest/enumflags2/struct.BitFlags.html#impl-TryFrom%3Cu32%3E-for-BitFlags%3CT%3E), but not `From<u32>`.

src/desktop/screencast.rs
```diff
@@ -183,11 +193,33 @@ impl Debug for Streams {
...
+        pipe_wire_node_id: u32,
+        position: Option<(i32, i32)>,
+        size: Option<(i32, i32)>,
+        source_type: Option<SourceType>,
+        mapping_id: Option<String>,
+    ) -> Self {
+        Self(
+            pipe_wire_node_id,
+            StreamProperties {
+                id: Some(pipe_wire_node_id.to_string()),
```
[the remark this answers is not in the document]
oSoMoN: You're right, my assumption that the `stream id == pipewire node id` is incorrect. I was tricked by the lack of documentation for that property in the backend API, but it is documented [in the frontend API](https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.portal.ScreenCast.html#org-freedesktop-portal-screencast-start):

> Opaque identifier. Will be unique for this stream and local to this session. Will persist with future sessions, if they are restored using a restore token. This property was added in version 4 of this interface. Optional.

src/backend/screencast.rs
```diff
@@ -0,0 +1,241 @@
...
+    fn available_source_types(&self) -> BitFlags<SourceType>;
+
+    fn available_cursor_mode(&self) -> BitFlags<CursorMode>;
+
+    async fn create_session(
+        &self,
+        handle: OwnedObjectPath,
+        session_handle: OwnedObjectPath,
+        app_id: Option<AppID>,
+        options: CreateSessionOptions,
```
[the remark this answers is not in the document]
oSoMoN: We could, but that wouldn't buy us much, just one fewer parameter. And as you point out, adding it back would result in an API break. In this regard, keeping the unused `options` parameter is future-proof.

I'll defer to your judgement though, if you have a strong preference for a simpler API I'm fine with removing the parameter.

src/backend/session.rs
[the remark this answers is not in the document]
bilelmoussaoui: Then we should probably make it use HandleToken here, as that is the unique bit of the session path anyways.

src/backend/session.rs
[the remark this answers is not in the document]
bilelmoussaoui: i would set it to 1 and assume that xdp-gnome doesn't set a value for it.

src/backend/session.rs
```diff
@@ -0,0 +1,62 @@
...
+
+    async fn close(
+        &self,
+        #[zbus(object_server)] server: &zbus::ObjectServer,
+        #[zbus(signal_emitter)] emitter: SignalEmitter<'_>,
+    ) -> zbus::fdo::Result<()> {
+        #[cfg(feature = "tracing")]
+        tracing::debug!("Closing session {:?}", self.path.as_str());
+        server.remove::<Self, _>(&self.path).await?;
+        let _ = Self::closed(&emitter).await;
```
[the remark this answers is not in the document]
bilelmoussaoui: right, that makes sense to me. As the closed signal is supposed to be emitted when something goes wrong, not when you intentionally call close yourself on it as you are aware of the session being closed after that.

src/backend/screencast.rs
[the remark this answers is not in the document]
bilelmoussaoui: Let us leave it like this for now.

backend-demo/src/screencast.rs
```diff
@@ -0,0 +1,113 @@
...
+
+    async fn select_sources(
+        &self,
+        session_handle: OwnedObjectPath,
+        _app_id: Option<AppID>,
+        _options: SelectSourcesOptions,
+    ) -> Result<SelectSourcesResponse> {
+        tracing::debug!("IN Screencast::select_sources(): {session_handle}");
+        let session = session_handle.to_string();
+        let sessions = self.sessions.lock().unwrap();
```
bilelmoussaoui, line 86: I know that we don't have support for remote desktop portal yet, but see https://github.com/bilelmoussaoui/ashpd/blob/master/src/desktop/remote_desktop.rs#L33-L37. Basically, the session existence check here, should also check for session existence for RemoteDesktop sessions... Which means, somehow sharing the state between the RemoteDesktop & ScreenCast impls (probably other cases like this in the future).

I don't have a solution to this right now, but we have to find a way to store all the sessions somewhere and make it easy to retrieve a session that corresponds to a HandleToken or to close it.

src/backend/screencast.rs
```diff
@@ -0,0 +1,241 @@
...
+                let path = ObjectPath::from_string_unchecked(session.path().as_str().to_string());
+                let response = imp
+                    .create_session(handle, path.into(), app_id.inner(), options)
+                    .await;
+                if response.is_ok() {
+                    #[cfg(feature = "tracing")]
+                    tracing::debug!("session created: {:?}", session);
+                    let path = session.path().as_str().to_string();
+                    self.cnx.object_server().at(path.clone(), session).await?;
+                    if let Some(name) = self.cnx.unique_name() {
```
bilelmoussaoui, line 164: The unique name should be retrieved from the handle token. if needed, we can add a method there for that.

src/backend/screencast.rs
```diff
@@ -0,0 +1,241 @@
...
+                        let proxy = SessionProxy::new(&self.cnx, name, path.clone()).await?;
+                        let mut closed_monitor = proxy.receive_closed().await?;
```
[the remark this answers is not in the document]
oSoMoN: You're right. It took me some thinking and experimentation, but I think I came up with a (more) elegant solution where interested backends implement a `SessionMonitor` trait (which declares a `session_closed` function), and the session constructor takes an optional instance of this trait.

src/backend/screencast.rs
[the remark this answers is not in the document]
oSoMoN: Got it. It sounded a bit counter-intuitive at first to deviate from the D-Bus API, but your reasoning about the only unique part of the path makes sense.

src/backend/screencast.rs
```diff
@@ -0,0 +1,241 @@
...
+        #[cfg(feature = "tracing")]
+        tracing::debug!("IN ScreencastInterface::create_session(): {handle}, {session_handle}, …");
+
+        Request::spawn(
+            "ScreenCast::CreateSession",
+            &self.cnx,
+            handle.clone(),
+            Arc::clone(&self.imp),
+            async move {
+                let session = Session::new(session_handle);
```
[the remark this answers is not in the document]
oSoMoN: Yes, that makes sense. I'm not sure how to best implement that right now and I'm wrapping up the week, I'll follow up on this one next week.

src/backend/session.rs
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    pub async fn serve(self, cnx: zbus::Connection) -> zbus::Result<bool> {
+        cnx.object_server().at(self.path.clone(), self).await
+    }
+}
+
+#[zbus::interface(
+    name = "org.freedesktop.impl.portal.Session",
+    proxy(gen_blocking = false)
```
bilelmoussaoui, line 33: the proxy part is no longer needed

backend-demo/src/screencast.rs
```diff
@@ -0,0 +1,117 @@
...
+        }
+        let streams = vec![StreamBuilder::new(42)
+            .source_type(SourceType::Monitor)
+            .build()];
+        Ok(StreamsBuilder::new(streams).build())
+    }
+}
+
+#[async_trait]
+impl SessionMonitor for Screencast {
```
bilelmoussaoui, line 103: What is the difference between SessionMonitor and a potentially named SessionImpl trait? Seems the same to me

backend-demo/src/screencast.rs
```diff
@@ -0,0 +1,117 @@
...
+        HandleToken,
+    },
+    enumflags2::BitFlags,
+    AppID, PortalError, WindowIdentifierType,
+};
+use async_trait::async_trait;
+
+#[derive(Default)]
+pub struct Screencast {
+    sessions: Arc<Mutex<Vec<String>>>,
```
bilelmoussaoui, line 24:
```suggestion
    sessions: Arc<Mutex<Vec<HandleToken>>>,
```

backend-demo/src/screencast.rs
oSoMoN: Indeed, no difference. The pattern of naming traits with an `Impl` suffix feels a bit weird to me, but consistency is worth it, so I'm renaming `SessionMonitor`to `SessionImpl`.

bilelmoussaoui: The naming comes from the GTK Rust bindings, where we define a trait named `$TypeImpl` for `Impl`ementing a custom `$Type` and I followed the same here, it is not great but good enough.

src/backend/screencast.rs
```diff
@@ -0,0 +1,241 @@
...
+                let path = ObjectPath::from_string_unchecked(session.path().as_str().to_string());
+                let response = imp
+                    .create_session(handle, path.into(), app_id.inner(), options)
+                    .await;
+                if response.is_ok() {
```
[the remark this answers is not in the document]
bilelmoussaoui: Please use a match and log the error as well.

src/backend/session.rs
```diff
@@ -0,0 +1,86 @@
...
+pub(crate) struct Session {
+    path: OwnedObjectPath,
+    monitor: Option<Arc<dyn SessionMonitor>>,
+}
+
+impl Session {
+    #[allow(dead_code)]
+    pub fn path(&self) -> ObjectPath<'_> {
+        self.path.as_ref()
+    }
```
bilelmoussaoui, line 20: If not used, then it should be removed

src/backend/session.rs
```diff
@@ -0,0 +1,86 @@
...
+
+    pub(crate) fn new(path: OwnedObjectPath, monitor: Option<Arc<dyn SessionMonitor>>) -> Self {
+        Self { path, monitor }
+    }
+
+    pub async fn serve(self, cnx: zbus::Connection) -> zbus::Result<bool> {
+        cnx.object_server().at(self.path.clone(), self).await
```
bilelmoussaoui, line 27:
```suggestion
        cnx.object_server().at(&self.path, self).await
```
no?

src/backend/session.rs
```diff
@@ -0,0 +1,86 @@
...
+    fn version(&self) -> u32 {
+        1
+    }
+
+    async fn close(
+        &self,
+        #[zbus(object_server)] server: &zbus::ObjectServer,
+    ) -> zbus::fdo::Result<()> {
+        #[cfg(feature = "tracing")]
+        tracing::debug!("Closing session {:?}", self.path.as_str());
```
bilelmoussaoui, line 46: do you need {:?} when using as_str instead of just {}?

src/desktop/screencast.rs
```diff
@@ -183,7 +183,38 @@ impl Debug for Streams {
...
+    pub fn new(streams: Vec<Stream>) -> Self {
+        Self {
+            streams: Streams {
+                streams,
+                restore_token: None,
+            },
+        }
+    }
+
+    /// Set the streams' optional restore token (TODO).
```
bilelmoussaoui, line 205: why is this todo?

backend-demo/src/screencast.rs
oSoMoN: Got it, thanks for the reference.

src/backend/session.rs
oSoMoN: No, indeed.

src/desktop/screencast.rs
oSoMoN: I can't remember why I left that TODO (sigh). I'm removing it.

src/backend/screencast.rs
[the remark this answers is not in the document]
oSoMoN: I refactored this logic with a `SessionManager` struct that the `ScreencastInterface` implementation refers to, so backend authors won't have to bother with low-level session management.

backend-demo/src/screencast.rs
oSoMoN: I made an attempt at addressing this future use case with a shared `SessionManager` struct that the backend builder instantiates and passes to the `ScreencastInterface` (and would pass to the `RemoteDesktopInterface` when it gets added).

Does that make sense?

src/backend/session.rs
```diff
@@ -0,0 +1,193 @@
...
+
+    pub fn error_notfound(&self, handle_token: &HandleToken) -> PortalError {
+        let errormsg = format!("Unknown session: `{handle_token}`");
+        #[cfg(feature = "tracing")]
+        tracing::error!("{}", errormsg.as_str());
+        PortalError::NotFound(errormsg)
+    }
+}
+
+impl Display for SessionManager {
```
bilelmoussaoui, line 182: No need for this, please drop it

src/backend/session.rs
```diff
@@ -0,0 +1,193 @@
...
+        PortalError::Exist(errormsg)
+    }
+
+    pub fn error_notfound(&self, handle_token: &HandleToken) -> PortalError {
+        let errormsg = format!("Unknown session: `{handle_token}`");
+        #[cfg(feature = "tracing")]
+        tracing::error!("{}", errormsg.as_str());
+        PortalError::NotFound(errormsg)
+    }
+}
```
bilelmoussaoui, line 180: I would make those private and add something like try_contains that returns a Err if it doesn't exist.

src/backend/session.rs
```diff
@@ -0,0 +1,193 @@
...
+        }
+        server.remove::<Self, _>(&self.path).await?;
+        Ok(())
+    }
+
+    #[zbus(signal)]
+    async fn closed(signal_emitter: &SignalEmitter<'_>) -> zbus::Result<()>;
+}
+
+impl Debug for Session {
```
bilelmoussaoui, line 104: Move this above

src/backend/screencast.rs
```diff
@@ -0,0 +1,261 @@
...
+    pub fn persist_mode(&self) -> Option<PersistMode> {
+        self.persist_mode
+    }
+}
+
+#[derive(SerializeDict, Type, Debug, Default)]
+#[zvariant(signature = "dict")]
+pub struct SelectSourcesResponse {}
+
+#[derive(DeserializeDict, Type, Debug, Default)]
```
bilelmoussaoui, line 61:
```suggestion
#[derive(DeserializeDict, Type, Debug)]
```

backend-demo/src/screencast.rs
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    async fn select_sources(
+        &self,
+        session_handle_token: HandleToken,
+        _app_id: Option<AppID>,
+        _options: SelectSourcesOptions,
+    ) -> Result<SelectSourcesResponse> {
+        tracing::debug!("IN Screencast::select_sources(): {session_handle_token}");
+        Ok(SelectSourcesResponse {})
```
bilelmoussaoui, line 58:
```suggestion
        Ok(SelectSourcesResponse::default())
```

backend-demo/src/screencast.rs
```diff
@@ -0,0 +1,82 @@
...
+        SourceType::Monitor | SourceType::Window
+    }
+
+    fn available_cursor_mode(&self) -> BitFlags<CursorMode> {
+        CursorMode::Hidden | CursorMode::Embedded | CursorMode::Metadata
+    }
+
+    async fn create_session(
+        &self,
+        _handle_token: HandleToken,
```
bilelmoussaoui, line 42: i would name all the _handle_token to just token and session_handle_token to just session_token

src/backend/session.rs
```diff
@@ -0,0 +1,193 @@
...
+    ) -> zbus::fdo::Result<()> {
+        #[cfg(feature = "tracing")]
+        tracing::debug!("SessionInterface::Close {}", self.path.as_str());
+        let handle_token = HandleToken::try_from(&self.path).unwrap();
+        {
+            let mut manager = self.manager.lock().unwrap();
+            let _ = manager.remove(&handle_token);
+        }
+        if let Some(monitor) = &self.monitor {
+            let _ = monitor.session_closed(handle_token).await;
```
bilelmoussaoui, line 94: I guess we said, no session_closed emitted when you Close is called?

bilelmoussaoui: Instead, it should be called if Session is dropped

src/backend/session.rs
```diff
@@ -0,0 +1,193 @@
...
+        } else {
+            Ok(())
+        }
+    }
+
+    /// Remove a session from the list of tracked sessions.
+    pub fn remove(&mut self, handle_token: &HandleToken) -> crate::backend::Result<()> {
+        let mut sessions = self.sessions.lock().unwrap();
+        if sessions.remove(handle_token).is_some() {
+            Ok(())
```
bilelmoussaoui, line 161: Here calling session_closed would make sense

bilelmoussaoui: but we also call Manager.remove from SessionInterface, to be resolved somehow.

src/backend/session.rs
```diff
@@ -0,0 +1,193 @@
...
+        if sessions.insert(handle_token.clone(), session).is_some() {
+            Err(self.error_exist(&handle_token))
+        } else {
+            Ok(())
+        }
+    }
+
+    /// Remove a session from the list of tracked sessions.
+    pub fn remove(&mut self, handle_token: &HandleToken) -> crate::backend::Result<()> {
+        let mut sessions = self.sessions.lock().unwrap();
```
bilelmoussaoui, line 159: No need for a mutex if you take &mut anyways...

bilelmoussaoui (commented):
few more comments, but getting there

backend-demo/src/screencast.rs
bilelmoussaoui: yep, seems like a decent approach till we have more constraints from other impls. Thanks!

src/backend/session.rs
```diff
@@ -0,0 +1,193 @@
+use std::{
+    collections::HashMap,
+    fmt::{Debug, Display, Formatter},
+    sync::{Arc, Mutex},
+};
+
+use async_trait::async_trait;
+use itertools::Itertools;
```
bilelmoussaoui, line 8: What this was added for? please drop it if possible

src/backend/session.rs
```diff
@@ -0,0 +1,193 @@
+use std::{
+    collections::HashMap,
+    fmt::{Debug, Display, Formatter},
```
bilelmoussaoui, line 3: all the codebase doesn't import much from std, especially not std::fmt::*

bilelmoussaoui:
Do you have a use case for this btw? just curious if this impl will be extensively tested / if you need it in a release soon-ish

src/backend/builder.rs
```diff
@@ -32,10 +34,12 @@ pub struct Builder {
     lockdown_impl: Option<Arc<dyn LockdownImpl>>,
     permission_store_impl: Option<Arc<dyn PermissionStoreImpl>>,
     print_impl: Option<Arc<dyn PrintImpl>>,
+    screencast_impl: Option<Arc<dyn ScreencastImpl>>,
     screenshot_impl: Option<Arc<dyn ScreenshotImpl>>,
     secret_impl: Option<Arc<dyn SecretImpl>>,
     settings_impl: Option<Arc<dyn SettingsImpl>>,
     wallpaper_impl: Option<Arc<dyn WallpaperImpl>>,
+    sessions: Arc<Mutex<SessionManager>>,
```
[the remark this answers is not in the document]
oSoMoN: Good point, the mutex that wraps the hashmap in `SessionManager` was a leftover from an early iteration, I'm removing it (and adding a comment to state that `SessionManager` is not thread-safe).

src/backend/session.rs
oSoMoN: It is needed for the debug log in `SessionManager::contains()`:
```rust
pub fn contains(&self, handle_token: &HandleToken) -> bool {
    #[cfg(feature = "tracing")]
    tracing::debug!("SessionManager::contains: tracked sessions: {}", &self);
    self.sessions.contains_key(handle_token)
}
```
For a debug output that looks like this:
```
DEBUG ashpd::backend::session: SessionManager::contains: tracked sessions: [`foo`, `baz`, `bar`]
```

bilelmoussaoui: Impl Debug then, not Display.

oSoMoN: Right, the debug output for contains will display a list that looks like this now:
```
[HandleToken("foo"), HandleToken("baz"), HandleToken("bar")]
```
which is less concise, but I guess is fine given that this is debug output anyway.
I'm getting rid of the additional itertools dependency by the way, which was required by the use of the [`format_with` method](https://docs.rs/itertools/0.12.1/itertools/trait.Itertools.html#method.format_with).

bilelmoussaoui: that is good enough :) it is just a debug output

src/backend/session.rs
oSoMoN: That was my initial preference, but there are cases where we want to error out when a session already exists (in `CreateSession`), and cases where we want to error out when a session *doesn't* exist (basically all other methods that act on a previously created session). So we would need two variants of `try_contains` with an opposite logic, and I don't know how to write this in a way that's not going to be confusing.

bilelmoussaoui: You can have a try_contains and contains, contains would call try_contains().is_ok() and that is pretty much it, no?

oSoMoN:
> Do you have a use case for this btw? just curious if this impl will be extensively tested / if you need it in a release soon-ish

I should probably have started with that… Apologies for not sharing more context upfront. I am contributing to SteamOS, and the [gamescope](https://github.com/ValveSoftware/gamescope/) compositor exposes a pipewire video node for screencasting. The missing bit is a ScreenCast portal backend that exposes that node, and I have a working proof of concept implemented in Rust with this branch.

Cargo.toml
```diff
@@ -36,6 +36,7 @@ gdk4wayland = { package = "gdk4-wayland", version = "0.9", optional = true }
 gdk4x11 = { package = "gdk4-x11", version = "0.9", optional = true }
```
bilelmoussaoui, line 36: before i forget, please move the "relates to #38" part of the commit header to the commit body

src/backend/session.rs
oSoMoN: Perhaps this is a bit confusing, but calling `monitor.session_closed` isn't emitting the `Closed` signal on the session interface, it merely notifies the backend implementer that a session was closed, in case they want to do some cleanup on their side.

bilelmoussaoui: ah then, please make sure it is properly documented

src/backend/session.rs
oSoMoN: I went for your suggestion, removed the extra check for existence in the `add()` method, and expanded a bit the documentation to make things more explicit.

src/backend/screencast.rs
```diff
@@ -0,0 +1,260 @@
...
+        app_id: MaybeAppID,
+        options: CreateSessionOptions,
+    ) -> Result<Response<CreateSessionResponse>> {
+        let session_token = HandleToken::try_from(&session_handle).unwrap();
+        {
+            let sessions = self.sessions.lock().unwrap();
+            if sessions.contains(&session_token) {
+                let errormsg = format!("A session with handle `{session_token}` already exists");
+                #[cfg(feature = "tracing")]
+                tracing::error!("ScreencastInterface::create_session: {}", errormsg.as_str());
```
bilelmoussaoui, line 151:
```suggestion
                tracing::error!("ScreencastInterface::create_session: {}", errormsg);
```
you don't need the as_str(), String impl Display.

src/backend/session.rs
```diff
@@ -0,0 +1,187 @@
...
+            let message = Self::notfound_msg(token);
+            #[cfg(feature = "tracing")]
+            tracing::error!("{}", message.as_str());
+            Err(PortalError::NotFound(message))
+        }
+    }
+
+    fn notfound_msg(token: &HandleToken) -> String {
+        format!("Unknown session: `{token}`")
+    }
```
bilelmoussaoui, line 180: not sure it is worth a method just for this little error message that is used twice ;-) i would drop it

bilelmoussaoui (approved):
otherwise lgtm

bilelmoussaoui:
thanks for your contribution! hopefully now, someone else motivated enough will help me finish the remaining backend support!

oSoMoN:
Thanks for the prompt and thorough review, and for the constructive feedback!

src/backend/screencast.rs
```diff
@@ -0,0 +1,260 @@
...
+#[zvariant(signature = "dict")]
+pub struct CreateSessionOptions {}
+
+#[derive(DeserializeDict, Type, Debug)]
+#[zvariant(signature = "dict")]
+pub struct SelectSourcesOptions {
+    types: Option<BitFlags<SourceType>>,
+    multiple: Option<bool>,
+    cursor_mode: Option<CursorMode>,
+    restore_token: Option<String>,
```
HigherOrderLogic: I was reading the spec and found that this field doesnt match the spec.

Shouldnt it be the following?

```suggestion
    restore_data: Option<(String, u32, Value<'static>)>,
```

bilelmoussaoui: Indeed.

HigherOrderLogic: Thanks for the quick reply.

PR #289, opened 2025-06-16
backend/app-chooser: Pass `HandleToken` for `update_choices`

arun-mani-j:
Please let me know if there are other request methods that needs similar changes :).

### DIFF
--- a/src/backend/app_chooser.rs
+++ b/src/backend/app_chooser.rs
@@ -8,7 +8,6 @@ use crate::{
         MaybeAppID, MaybeWindowIdentifier,
     },
     desktop::{HandleToken, Response},
-    zbus::object_server::{InterfaceRef, ObjectServer},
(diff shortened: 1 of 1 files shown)


src/backend/app_chooser.rs
```diff
@@ -143,15 +142,14 @@ impl AppChooserInterface {
...
     async fn update_choices(
         &self,
-        #[zbus(object_server)] server: &ObjectServer,
         handle: OwnedObjectPath,
         choices: Vec<AppID>,
     ) -> Result<(), PortalError> {
         #[cfg(feature = "tracing")]
         tracing::debug!("AppChooser::UpdateChoices");
 
-        let iface_ref = server.interface::<_, Request>(handle).await?;
```
bilelmoussaoui, line 153: Something that might be worth adding, is two methods to HandleToken that gives you an InterfaceRef<Request> or InterfaceRef<Session>, taking care of the conversion back from a HandleToken to an ObjectPath, what do you think?

PR #292, opened 2025-06-23
Allow passing a custom future spawner to backend builder

### LINKED ISSUE #278: Backend without Tokio
There are 2 usages of tokio within `backend`, both in `request.rs`:

- `tokio::sync::Mutex`, this is easily replaced with `futures_util::lock::Mutex` (with a `cfg(not(feature = "tokio"))` to keep implementation the same when possible)
- `tokio::spawn`, afaik the possible options are adding more specific runtimes as features or taking a `impl futures_util::task::Spawn` in `backend::Builder::new()`

This will be a breaking change for projects using `backend` without explicitly enabling `tokio` (like `backend-demo`).

P.S. Would it make sense to allow the portal `*Impl`s to access the underlying `zbus::connection::Connection`? This seems useful for e.g. delegating to internal dbus apis like gnome-shell, my use case is proxying to another portal implementation.

happenslol:
Closes #278.

I've tested this using both tokio and a custom executor (gpui).

I'm not super happy with this, as it potentially introduces a bit of runtime overhead since the executor now has to be passed around and cloned. It would be possible to add feature flags to every place it's used and avoid that, but that adds a lot of noise to a lot of places - I'd be happy to do that though, if you're fine with that.

That way, it would also be possible to make the executor a required parameter on the builder, right now it errors at runtime when building the backend if you didn't pass a spawner and have `tokio` disabled.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,7 +15,7 @@ rust-version = "1.75"
 async-std = ["zbus/async-io", "dep:async-fs", "dep:async-net", "futures-util/io"]
 default = ["tokio"]
 
-backend = ["async-trait", "tokio"]
+backend = ["async-trait"]
 
 gtk4 = ["gtk4_x11", "gtk4_wayland"]
 gtk4_wayland = ["gdk4wayland", "glib", "dep:gtk4"]
--- a/backend-demo/Cargo.toml
+++ b/backend-demo/Cargo.toml
@@ -15,5 +15,5 @@ zbus = "5.0"
 
 [dependencies.ashpd]
 path = "../"
-features = ["backend", "tracing"]
+features = ["backend", "tracing", "tokio"]
 default-features = false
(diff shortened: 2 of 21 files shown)


happenslol:
I don't think the overhead is impactful to be honest, but it could be cleaner. Maybe a feature to specifically enable custom spawners for the backend, and then it's always required? Then I'd make the argument dependant on the flag everywhere.

I'm only unhappy with the runtime error when no spawner is passed, to be honest.

bilelmoussaoui:
> I'm only unhappy with the runtime error when no spawner is passed, to be honest.

Tbh, the current approach seems good enough, given the panic at runtime if ashpd is built without tokio.

happenslol:
All done, that only leaves the one todo I added. I'm fine with leaving that as-is for now, or just adding an error log with tracing?

bilelmoussaoui:
> All done, that only leaves the one todo I added. I'm fine with leaving that as-is for now, or just adding an error log with tracing?

Let us go with adding an error log with tracing.

happenslol:
Fixed, the backend demo needs to enable tokio now since the `backend` feature no longer does so.

happenslol:
My bad, formatting fixed

bilelmoussaoui (approved):

PR #298, opened 2025-08-07
ashpd-demo: Add snap build

3v1n0:
ashpd is the the reference tool to test portals these days, and portals are supported by (and supports) both flatpaks and snaps, so it would be relevant to have an official ashpd build to verify the snaps behaviors with portals.

Thus, adding a snap build here.

I would prefer to maintain it upstream instead of having a build-only fork, but please @bilelmoussaoui, let me know if interested and in case if you want to register the snap name under your ownership in the store, and if you care about setting up CI for automatic builds.

### DIFF
--- a/ashpd-demo/build-aux/snap/snapcraft.yaml
+++ b/ashpd-demo/build-aux/snap/snapcraft.yaml
@@ -1,0 +1,98 @@
+name: ashpd-demo
+base: core24
+adopt-info: ashpd-demo
+
+grade: stable
+confinement: strict
+
+slots:
+  ashpd-demo:
+    interface: dbus
+    bus: session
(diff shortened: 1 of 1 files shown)


bilelmoussaoui:
As the demo is indeed useful, I am considering merging your patch. Can this recipe be built on CI to ensure the demo keeps working? does it build from main?

3v1n0:
> Can this recipe be built on CI to ensure the demo keeps working?

Sure, I can handle that. We can also automate uploads to the store, but we may need you to take control of the name in the snap store or ask @kenvandine (the current owner) to pass it to someone else or manage it.

> does it build from main?

Yes, since the name in the snap world normally does not change for "edge" uploads, but but we could even make it build the devel version.

PR #303, opened 2025-08-28
Switch from `rand` to `fastrand`

jakobhellermann:
Currently, `ashpd` uses both `fastrand` through `zbus->futures-lite->fastrand`, and `rand`.

Rand supports various abstractions for generators and distributions, but `ashpd` only has a very basic use case of generating a simple token.

By switching to `fastrand` we can cut a few dependencies:

```
    Removing ppv-lite86 v0.2.21
    Removing rand v0.9.2
    Removing rand_chacha v0.9.0
    Removing rand_core v0.9.3
    Removing zerocopy v0.8.26
    Removing zerocopy-derive v0.8.26
```

Whether this actually saves time depends on how the remaining build looks like, but it doesn't hurt. On a clean release build of `ashpd` there isn't really a measurable speedup, but there's at least potential to skip the `1s` build (and build script) of `zerocopy`.

image

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -99,9 +99,9 @@ dependencies = [
 
 [[package]]
 name = "async-executor"
-version = "1.13.3"
+version = "1.13.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


jakobhellermann:
Actually, does the `HandleToken` have to be cryptographically secure? if so, this shouldn't be merged, `ThreadRng` is better in that regard compared to `fastrand`.

PR #305, opened 2025-10-16
Add set_session_connection function for custom zbus connections

technobaboo:
I'm making a UI toolkit that establishes its own connection through zbus so I want to simplify it and use the connection I already have, idk if this is the best way to do that but I figured it fit most in line with what you had already...

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -33,6 +33,7 @@ mod file_path;
 pub use self::file_path::FilePath;
 
 mod proxy;
+pub use proxy::set_session_connection;
 
 #[cfg(feature = "backend")]
 #[cfg_attr(docsrs, doc(cfg(feature = "backend")))]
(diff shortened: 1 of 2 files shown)


bilelmoussaoui:
Which APIs you are using the UI toolkit?

bilelmoussaoui:
> it's early days for my toolkit, so i use accent color (and keep the changed stream open for live updates) and am planning to use the file chooser and documents portals soon.
> 
> When debugging my toolkit in d-spy i noticed 2 connections opened and stayed open, 1 of which was just empty... i noticed it in cosmic apps too and figured out it was ashpd opening a second connection that must remain open to receive accent color change notifications.

I rather we expose another constructor to the Settings (proxy) that is basically with_connection that constructs the proxy using your own connection. We would have with time to expose more proxies constructors instead of your approach. Mind updating your patch?

bilelmoussaoui:
I did that in https://github.com/bilelmoussaoui/ashpd/commit/1df8823beb7636b0d6a62b1f66a6cb84357b15ee, supporting all the interfaces so you should be good.

PR #307, opened 2025-10-27
demo: window: Port to AdwViewSwitcherSidebar

A6GibKm:
Lightly tested.

### DIFF
--- a/ashpd-demo/src/window.rs
+++ b/ashpd-demo/src/window.rs
@@ -1,9 +1,7 @@
-use adw::subclass::prelude::*;
-use gettextrs::gettext;
+use adw::{prelude::*, subclass::prelude::*};
 use gtk::{
     gio,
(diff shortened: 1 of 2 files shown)


ashpd-demo/src/window.rs
```diff
@@ -134,40 +129,28 @@ mod imp {
...
-                    .add_named(&BackgroundPage::default(), Some("background"));
-            } else {
-                self.sidebar
-                    .insert(&SidebarRow::new(&gettext("Device"), "device"), 2);
-                self.stack.add_named(&DevicePage::default(), Some("device"));
-
-                self.sidebar.insert(
-                    &SidebarRow::new(&gettext("Network Monitor"), "network_monitor"),
-                    7,
+                self.stack.add_titled(
```
bilelmoussaoui: don't we lose the order here?

PR #321, opened 2025-12-25
Move window identifiers and similar to a new ashpd-handles crate

A6GibKm:
TODO:
 - [ ] test all combinations of features of ashpd-handles. I think there is a preexisting issue where raw_handles+gtk4_wayland (or gtk4_x11) fails to build, but raw_handles,gtk4 works.
 - [ ] test the demo (cannot build right now cus Builder, and because my toolbox does not have new enough deps)
 - [x] Write into the commit why this is done, mostly to simplify the feature matrix testing, allow using the main branch of gtk-rs-core in a follow up (so that users are not blocked from using gtk-rs' main branch), and most importantly decouple the releases of ashpd from gtk-rs, however since we still have the glib feature for enums and a gtk4 feature for a gdk::RGBA this is not a complete solution.

Depends on https://github.com/bilelmoussaoui/ashpd/pull/320.

### DIFF
--- a/.github/workflows/flatpak-ci.yml
+++ b/.github/workflows/flatpak-ci.yml
@@ -2,7 +2,7 @@ on:
   push:
     branches: [master]
   pull_request:
-name: Client Demo FLatpak
+name: Client Demo Flatpak
 jobs:
   flatpak:
     name: "Flatpak"


bilelmoussaoui:
can you rebase?

A6GibKm:
ci passes now. this is prob not ready, but good enough for a review.

PR #336, opened 2026-01-19
Update to gtk-rs-core 0.22.0-alpha

A6GibKm:


client/Cargo.toml
```diff
@@ -80,8 +80,8 @@ flatpak = ["dep:serde_repr"]
 documents = ["dep:serde_repr"]
 
 gtk4 = ["gtk4_x11", "gtk4_wayland"]
-gtk4_wayland = ["gdk4wayland", "glib", "dep:gtk4"]
-gtk4_x11 = ["gdk4x11", "glib", "dep:gtk4"]
+gtk4_wayland = ["gdk4wayland", "gio-unix", "glib", "dep:gtk4"]
+gtk4_x11 = ["gdk4x11", "gio-unix", "glib", "dep:gtk4"]
```
bilelmoussaoui: not sure pulling yet another crate just for this is useful... i rather we drop that helper from AppId

PR #347, opened 2026-02-02
backend/request: Match `PortalError` to `Response`

arun-mani-j:
The portal docs say that a response must return a number that signals whether the request succeeded or not - https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.portal.Request.html#org-freedesktop-portal-request-response.

In `ashpd`, every backend method returns a `Result<T, PortalError>`. So whenever backend returns a `PortalError` for cancelled or other type of errors, `ashpd` throws them as a DBus error instead of returning a number `1` (for cancelled request) or `2` (for other errors).

For a simple example:
```diff
diff --git a/demo/backend/src/wallpaper.rs b/demo/backend/src/wallpaper.rs
index 7eb89c0..62404d1 100644
--- a/demo/backend/src/wallpaper.rs
+++ b/demo/backend/src/wallpaper.rs
@@ -29,6 +29,6 @@ impl WallpaperImpl for Wallpaper {
         _uri: url::Url,
         _options: WallpaperOptions,
     ) -> Result<()> {
-        Ok(())
+        Err(ashpd::PortalError::Cancelled("user cancelled the request".to_string()))
     }
 }
```

Now run the backend:
```sh
$ cargo run
```

Make a request to GNOME (or any portal implementation that does not use `ashpd`) and when the dialog is shown, cancel it:
```sh
$ busctl --user call org.freedesktop.impl.portal.desktop.gnome /org/freedesktop/portal/desktop org.freedesktop.impl.portal.Wallpaper SetWallpaperURI osssa\{sv\} "/foo" "com.foo.x" "spam" "file:///foo/bar" 1 "show-preview" b true
u 1
```

Now do the same with `ashpd`:
```sh
$ busctl --user call org.freedesktop.impl.portal.desktop.ashpd-backend-demo /org/freedesktop/portal/desktop org.freedesktop.impl.portal.Wallpaper SetWallpaperURI osssa\{sv\} "/foo" "com.foo.x" "spam" "file:///foo/bar" 1 "show-preview" b true
Call failed: user cancelled the request
```

After this commit, `ashpd` also returns proper numeric response:
```sh
$ busctl --user call org.freedesktop.impl.portal.desktop.ashpd-backend-demo /org/freedesktop/portal/desktop org.freedesktop.impl.portal.Wallpaper SetWallpaperURI osssa\{sv\} "/foo" "com.foo.x" "spam" "file:///foo/bar" 1 "show-preview" b true
u 1
```

The above fix is from my understanding of the issue, but I'm not sure if there is a specific reason why `ashpd` took the `PortalError` route. Checking https://github.com/flatpak/xdg-desktop-portal/blob/1.20.0/src/xdp-utils.h#L88-L96 (which the docs for `PortalError` points to), it seems the error enum defined in `xdp-utils.c` is also unused.

So I'm not sure why `PortalError` was used and happy to update the PR following suggestions. :)

### DIFF
--- a/client/src/backend/request.rs
+++ b/client/src/backend/request.rs
@@ -8,7 +8,10 @@ use futures_util::{
 };
 use zbus::zvariant::{ObjectPath, OwnedObjectPath};
 
-use crate::desktop::{HandleToken, Response};
+use crate::{
+    PortalError,
+    desktop::{HandleToken, Response},
+};
 
 #[async_trait]
 pub trait RequestImpl: Send + Sync {
@@ -63,18 +66,27 @@ impl Request {
         );
         server.at(&path, request).await?;
 
-        let response = fut.await;
+        let result = fut.await;
 
         #[cfg(feature = "tracing")]
-        tracing::debug!("{_method} returned {:#?}", response);
+        tracing::debug!("{_method} returned {:#?}", result);
         #[cfg(feature = "tracing")]
         tracing::debug!("Releasing request {:?}", path.as_str());
         server.remove::<Self, _>(&path).await?;
 
-        Ok(match response {
+        let response = match result {
             Err(_) => Response::cancelled(),
-            Ok(response) => Response::ok(response?),
-        })
+            Ok(result) => match result {
+                Ok(results) => Response::ok(results),
+                Err(error) => match error {
+                    PortalError::Cancelled(_) => Response::cancelled(),
+                    other => {
+                        return Err(other);
+                    }
+                },
+            },
+        };
+        Ok(response)
     }
 
     pub(crate) fn new(


bilelmoussaoui (approved):

arun-mani-j:
> > I personally think it is good to have a well described enum like PortalError than abstracting everything behind a vague other. Also the docs for other say "The user interaction was ended in some other way". Not sure how standard wants errors to be expressed clearly through this methodology. :)
> 
> I guess the issue is about when to propagate the error as PortalError (failing on dbus method call) and when to return Other? I am not sure what would be the appropriate way but i would just always propagate the error and ignore Other variant from ashpd backend side for now...

Got it. Currently if we the `impl`s return any error other than `PortalError::Cancelled`, we send `Response::other()`. Shall I change it to throw a DBus error instead?

arun-mani-j:
Hi, is it possible to backport this to stable release? 🙏

bilelmoussaoui:
cherry-picked into the new release.

bilelmoussaoui:
actually it is already in a 0.13 release?

arun-mani-j:
Thanks! I think the last time I checked, it was not part of release or may I did not check properly. All good then, thanks.

PR #352, opened 2026-02-11
add from_xdg_foreign_exported_v2 to window identifier

feschber:
this enables one to use an externally exported handle directly (see https://github.com/feschber/lan-mouse/pull/384)


bilelmoussaoui:
I was going to add the _raw part but that seemed a bit too long of a name. For the unsafe part, not sure why it would have to be unsafe?

PR #359, opened 2026-02-17
fix: Session::Closed signal being silently discarded

cadubentzen:
The signal was being silently discarded because ashpd was trying to deserialize to () while the signal is:

```
Closed (
  details a{sv}
)
```


bilelmoussaoui:
thanks a lot!

PR #362, opened 2026-02-23
clipboard: Add missing mime types setter to SetSelectionOptions

whot:
Fixes: effaceb359ca ("Always use an Options struct")


bilelmoussaoui:
Thanks! Will backport later today

PR #365, opened 2026-02-23
inputcapture: Fix create_session2 invocation

whot:
CreateSession2 only takes the session handle token, nothing else. And it's a normal method call without a Request, it immediately returns.

Fixes: 16eb6d2d136b ("input-capture: Add new methods")

### DIFF
--- a/client/src/desktop/input_capture.rs
+++ b/client/src/desktop/input_capture.rs
@@ -350,6 +350,21 @@ struct CreateSessionResponse {
 
 #[derive(Debug, Serialize, Type, Default)]
 #[zvariant(signature = "dict")]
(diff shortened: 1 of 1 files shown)


client/src/desktop/input_capture.rs
```diff
@@ -735,20 +742,15 @@ impl InputCapture {
     #[doc(alias = "CreateSession2")]
     pub async fn create_session2(
         &self,
-        identifier: Option<&WindowIdentifier>,
         options: crate::desktop::CreateSessionOptions,
```
bilelmoussaoui: Then the usage of the common options type doesn't make sense...

PR #368, opened 2026-02-26
inputcapture: Move the barrier position into a struct

whot:
This makes code less ambiguous because we no longer carry tuples around without seeing what the four values represent.

This lets us re-use the same struct in the demo code too then.

### DIFF
--- a/client/src/desktop/input_capture.rs
+++ b/client/src/desktop/input_capture.rs
@@ -84,7 +84,7 @@
 //!
 //! ```rust,no_run
 //! use ashpd::desktop::input_capture::{
-//!     Barrier, BarrierID, Capabilities, CreateSessionOptions, InputCapture,
+//!     Barrier, BarrierID, BarrierPosition, Capabilities, CreateSessionOptions, InputCapture,
 //! };
 //!
 //! #[allow(unused)]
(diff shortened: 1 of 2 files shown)


whot:
on that note: would you be ok with a pre-commit config? I keep getting caught out here because all other repos I work on do cargo fmt/clippy on commit :)

demo/client/src/portals/desktop/input_capture.rs
```diff
@@ -97,41 +97,21 @@ glib::wrapper! {
...
-            } else {
-                "Right".to_string()
-            }
-        } else if self.y1 == self.y2 {
-            if self.y1 == 0 {
-                "Top".to_string()
-            } else {
-                "Bottom".to_string()
-            }
+pub fn edge_name(position: &BarrierPosition) -> String {
```
bilelmoussaoui:
```suggestion
pub fn edge_name(position: &BarrierPosition) -> &'static str {
```
while you are at it

client/src/desktop/input_capture.rs
```diff
@@ -658,24 +658,87 @@ impl Zones {
 /// A barrier ID.
 pub type BarrierID = NonZeroU32;
 
+/// Position of a barrier defined by two points (x1, y1) and (x2, y2).
+///
+/// Barriers are typically placed along screen edges.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, Type)]
+#[zvariant(signature = "(iiii)")]
+pub struct BarrierPosition {
+    /// x coordinate of the first point
```
bilelmoussaoui: We tend to expose getters and setters instead

bilelmoussaoui: Or just getters in this case

PR #395, opened 2026-03-15
Lower rust-version from 1.92 to 1.87

### LINKED ISSUE #394: Consider lowering rust-version to 1.87 and adopting an MSRV policy
ashpd 0.13.0 declares rust-version = "1.92" and edition = "2024", but the source code does not use any language features beyond Rust 1.85. There are no unsafe functions (so unsafe_op_in_unsafe_fn edition semantics are irrelevant), no gen keyword usage, and no edition-2024-specific syntax.

The transitive dependency driving the real MSRV floor is zbus 5.14 at 1.87. The ashpd code itself has no features requiring anything above that.

Since edition 2024 was stabilized in Rust 1.85, the edition = "2024" setting is fully compatible with rust-version = "1.87". These are not exclusive.

This matters for Linux distribution packaging:

| Distribution | rustc version | ashpd 0.13 buildable? |
|---|---|---|
| Debian 13 (trixie, current stable) | 1.85.0 | No (but only 2 minor versions from 1.87) |
| Fedora 43 | 1.90.0 | No |
| RHEL 9 | 1.88.0 | No |
| openSUSE Tumbleweed | 1.92+ | Yes |
| Arch Linux | latest | Yes |

ashpd is the standard Rust interface to XDG Portals. Any Rust application using portals on these distributions is pinned to 0.12.x because 0.13.x declares an MSRV above what the distribution provides, despite the code not requiring it.

Lowering rust-version to 1.87 (matching the actual zbus dependency floor) would unblock Fedora, RHEL, and most distributions without any code changes.

I would also suggest adopting a documented MSRV policy. We raised the same question with IronRDP (Devolutions/IronRDP#1156), where they adopted rust-version in workspace Cargo.toml and documented their policy as part of PR #1153. A simple policy like "MSRV follows the oldest Rust version shipped by any current stable Debian, Fedora, or RHEL release" gives downstream packagers a clear contract without constraining development velocity.

glamberson:
Closes #394

The ashpd source code does not use any language features beyond Rust 1.85. The actual MSRV floor is driven by zbus 5.14 at 1.87. Edition 2024 was stabilized in Rust 1.85 and is fully compatible with rust-version 1.87. These are not exclusive.

This one-line change unblocks distribution packaging on Fedora 43 (rustc 1.90), RHEL 9 (rustc 1.88), and brings Debian 13 (rustc 1.85) within two minor versions which are backported adn therefore doable..

The rust-version was bumped from 1.75 to 1.92 in PR #320 alongside the workspace restructuring, but the code itself does not require anything above the zbus dependency floor.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,7 +12,6 @@ keywords = ["portal", "flatpak", "xdg", "desktop", "dbus"]
 license = "MIT"
 repository = "https://github.com/bilelmoussaoui/ashpd"
 version = "0.13.0"
-rust-version = "1.92"
 
 [workspace.dependencies]
 zvariant = { version = "5.8", default-features = false, features = [
--- a/client/Cargo.toml
+++ b/client/Cargo.toml
@@ -8,7 +8,7 @@ license.workspace = true
 name = "ashpd"
 repository.workspace = true
 version.workspace = true
-rust-version.workspace = true
+rust-version = "1.87"
 
 [features]
 async-io = ["zbus/async-io", "futures-util/io"]
--- a/demo/backend/Cargo.toml
+++ b/demo/backend/Cargo.toml
@@ -2,6 +2,7 @@
 name = "ashpd-backend-demo"
 version = "0.1.0"
 edition.workspace = true
+rust-version = "1.87"
 license.workspace = true
 
 [dependencies]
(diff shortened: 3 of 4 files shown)


glamberson:
Yes, that would be very helpful! We can start the migration to 0.13.x as soon as it lands on crates.io.

PR #420, opened 2026-04-04
app-chooser: Allow to build DesktopID from AppID

agx:
This makes it simple to construct a desktop-id from an app-id.

Maybe I'm missing something but I think that is needed to return `DesktopId` in a backend implementation.

### DIFF
--- a/client/src/backend/app_chooser.rs
+++ b/client/src/backend/app_chooser.rs
@@ -28,6 +28,12 @@ impl DesktopID {
     }
 }
 
+impl From<AppID> for DesktopID {
+    fn from(value: AppID) -> Self {
+        Self(Ok(value))
+    }
+}
+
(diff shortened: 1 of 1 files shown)


agx:
Thanks for merging! Is there a chance that this makes it into a 0.13.10, this would allow us to switch back with phosh's portal to a release.

bilelmoussaoui:
Will do a release next week

agx:
Thanks a lot!

agx:
Gentle ping? A release including this and https://github.com/bilelmoussaoui/ashpd/pull/427 would allow us to use crates rather than git again for phosh's portal. If I can help with anything, let me know.

bilelmoussaoui:
0.13.10 is out.

agx:
Thank you!

PR #433, opened 2026-04-22
tree: Use `DesktopID` for all backends

### LINKED ISSUE #432: Backend calls by apps with invalid app IDs are rejected
Because of https://github.com/bilelmoussaoui/ashpd/commit/f9333f2e720fb1b0d8485b57887a1f8252c11c10, apps like Firefox's requests fail with invalid app ID error.

Previously, the request gets to the implementation with `None` for app ID, but this no longer works.

I think bringing back the `MaybeAppID` or making `AppID` fallback to `None` for invalid app IDs would be good. Or perhaps this can be solved in other way?

arun-mani-j:
`AppID` is strict in accepting only valid application IDs. So requests from apps like Firefox (which has the ID `firefox`) will fail with `Invalid app ID` error.

So let us use `DesktopID` to indicate this scenario.

Closes: https://github.com/bilelmoussaoui/ashpd/issues/432

### DIFF
--- a/client/src/app_id.rs
+++ b/client/src/app_id.rs
@@ -1,6 +1,6 @@
 use std::{ops::Deref, str::FromStr};
 
-use serde::{Deserialize, Serialize};
+use serde::{Deserialize, Deserializer, Serialize};
 use zbus::zvariant::{self, Basic, Type};
 
 /// The application ID.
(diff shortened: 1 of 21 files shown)


bilelmoussaoui:
I wouldn't use DesktopId directly though, i rather we duplicate the type to avoid a misconception that it is actually a desktop id when it is not...

bilelmoussaoui:
Or maybe we should rename DesktopId to something else

arun-mani-j:
> Or maybe we should rename DesktopId to something else

May be we can name it to `MaybeAppID`?

I'm not sure of duplication because we introduced `DesktopID` to handle the exactly same problem in `AppChooser`. So I feel reusing the same type both places is fine.

bilelmoussaoui:
> May be we can name it to MaybeAppID?

Sounds better indeed

arun-mani-j:
> > May be we can name it to MaybeAppID?
> 
> Sounds better indeed

Thanks, done.

client/src/app_id.rs
```diff
@@ -87,6 +87,77 @@ impl<'de> Deserialize<'de> for AppID {
     }
 }
 
+/// The desktop ID of an application.
```
bilelmoussaoui: """
The application ID or the desktop file name without `.desktop` suffix.

As unsandboxed applications might not be defining a strict application ID, this type allows the backend implementation to fallback to the inner string without causing an error and manually handle those apps.
"""
Something along this line for the comment? could use some language fixes though :)

arun-mani-j: Right, done :).

bilelmoussaoui:
This needs fixing so the CI passes and i would make the type available only if the backend feature is enabled

arun-mani-j:
> This needs fixing so the CI passes and i would make the type available only if the backend feature is enabled

Done.

arun-mani-j:
Can a new release of ASHPD be done with this change? It will help us in unbreaking the Phosh portal - https://gitlab.gnome.org/World/Phosh/xdg-desktop-portal-phosh/-/work_items/27. :sweat_smile:

bilelmoussaoui:
Backported to 0.13.11

arun-mani-j:
Thanks!

PR #479, opened 2026-08-02
backend/screencast: consistently provide token

HigherOrderLogic:
Consistently provide the request token so that backend can close respective request, instead of needing to use some other mechanism of passing the request token around.

### DIFF
--- a/client/src/backend/screencast.rs
+++ b/client/src/backend/screencast.rs
@@ -43,6 +43,7 @@ pub trait ScreencastImpl: RequestImpl + SessionImpl {
     #[doc(alias = "SelectSources")]
     async fn select_sources(
         &self,
+        token: HandleToken,
         session_token: HandleToken,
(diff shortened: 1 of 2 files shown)


bilelmoussaoui:
If we are doing this for screen cast, we should make sure other ifaces are doing the same as well

HigherOrderLogic:
Do you mean passing all the tokens from the request to the handle? Cause some interface (e.g. [this](https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.impl.portal.RemoteDesktop.html#org-freedesktop-impl-portal-remotedesktop-notifypointermotion)) only provides one of the two.

bilelmoussaoui:
the changes are lgtm, but please rebase and make the related CI bits happy ;)

HigherOrderLogic:
The CI failures are unrelated.

PR #481, opened 2026-08-19
Add pipewire_serial support to the ScreenCast portal

glamberson:
This adds support for the pipewire_serial property that GNOME's
ScreenCast portal interface gained at version 6.

The node id returned by pipe_wire_node_id can be reused after the
underlying PipeWire node is destroyed, so a client that only tracks
the node id can misidentify a stream across monitor hotplug,
resolution changes, or suspend and resume. The pipewire_serial value
is a monotonically increasing identifier that is never reused, and
combined with PW_KEY_TARGET_OBJECT it is the way GNOME recommends
connecting to the stream safely.

I added the field to StreamProperties, a pipewire_serial getter on
Stream, and a pipewire_serial setter on StreamBuilder, matching the
existing mapping_id field as a precedent, including the same
serialization attributes and the same builder pattern. I did not add
it to the Debug impl for Stream, since mapping_id is not in that impl
either and I wanted to keep the diff consistent with the existing
field rather than fix an unrelated inconsistency.

I also bumped the version property in the backend ScreencastImpl
interface from 5 to 6, since a backend author who implements this
trait needs the version they advertise to reflect that
pipewire_serial is now available to populate.

Relevant portal interface documentation:
https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.portal.ScreenCast.html

### DIFF
--- a/client/src/backend/screencast.rs
+++ b/client/src/backend/screencast.rs
@@ -103,7 +103,7 @@ impl ScreencastInterface {
 
     #[zbus(property(emits_changed_signal = "const"), name = "version")]
     fn version(&self) -> u32 {
-        5
+        6
     }
 
     #[zbus(name = "CreateSession")]
--- a/client/src/desktop/screencast.rs
+++ b/client/src/desktop/screencast.rs
@@ -33,6 +33,7 @@
 //!         println!("node id: {}", stream.pipe_wire_node_id());
 //!         println!("size: {:?}", stream.size());
 //!         println!("position: {:?}", stream.position());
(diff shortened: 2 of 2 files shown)


glamberson:
@bilelmoussaoui just flagging this one in case it slipped by, no rush if you're deep in other things. This adds `pipewire_serial` deserialization to `StreamProperties` for the ScreenCast portal (the field GNOME's v6 portal added, closing a gap where our own portal backend was already emitting it but ashpd had nowhere to put it). Happy to adjust anything about the shape if you'd rather it land differently.

PR #482, opened 2026-09-01
backend: authenticate portal frontend callers

domenkozar:
## Summary

- add a reusable `CallerAuthorization` policy for backend D-Bus interfaces
- require the current owner of `org.freedesktop.portal.Desktop` for frontend-facing portal methods by default
- carry the policy into backend `Request::Close` and `Session::Close` objects
- authorize Lockdown property access with a separately configurable policy that inherits the frontend policy by default
- give PermissionStore a separately configurable policy that allows all callers by default because the document portal and PipeWire permission manager are legitimate non-Desktop callers

## Motivation

Backend methods trust frontend-supplied application IDs, handles, and options. Without authenticating the frontend, another process on the session bus can call implementation interfaces directly, spoof an application ID, manipulate an in-flight request, or close a session.

Most implementation calls originate from `org.freedesktop.portal.Desktop`, but Lockdown is intended for multiple portal frontends and PermissionStore has additional service callers. The policy is therefore shared while those two interfaces remain independently configurable.

## Testing

Run in ashpd's CI container with Rust 1.92.0:

- `cargo test --workspace --locked --features gtk4,pipewire,wayland,raw_handle,tracing,backend,frontend`
- `cargo clippy --workspace --features gtk4,pipewire,wayland,raw_handle,tracing,backend,frontend -- -D warnings`
- `cargo check --workspace --locked --features gtk4,pipewire,wayland,raw_handle,tracing,backend,frontend`
- `cargo fmt --all -- --check`
- `RUSTDOCFLAGS='-D warnings' cargo doc -p ashpd --all-features --no-deps`
- feature-minimal checks for `async-io,backend` and `tokio,backend,documents`

### DIFF
--- a/client/src/backend/access.rs
+++ b/client/src/backend/access.rs
@@ -2,11 +2,13 @@ use std::sync::Arc;
 
 use async_trait::async_trait;
 use serde::{Deserialize, Serialize};
+use zbus::message::Header;
 
 use crate::{
     MaybeAppID, WindowIdentifierType,
     backend::{
         Result,
+        caller::CallerAuthorization,
         request::{Request, RequestImpl},
     },
     desktop::{HandleToken, Icon, request::Response},
@@ -117,6 +119,7 @@ pub(crate) struct AccessInterface {
     imp: Arc<dyn AccessImpl>,
     spawn: Arc<dyn futures_util::task::Spawn + Send + Sync>,
     cnx: zbus::Connection,
+    caller_authorization: Arc<CallerAuthorization>,
 }
 
 impl AccessInterface {
@@ -124,8 +127,14 @@ impl AccessInterface {
         imp: Arc<dyn AccessImpl>,
         cnx: zbus::Connection,
         spawn: Arc<dyn futures_util::task::Spawn + Send + Sync>,
+        caller_authorization: Arc<CallerAuthorization>,
     ) -> Self {
-        Self { imp, cnx, spawn }
+        Self {
+            imp,
+            cnx,
+            spawn,
+            caller_authorization,
+        }
     }
 }
 
@@ -147,12 +156,17 @@ impl AccessInterface {
         subtitle: String,
         body: String,
         options: AccessOptions,
+        #[zbus(header)] header: Header<'_>,
     ) -> Result<Response<AccessResponse>> {
+        self.caller_authorization
+            .authorize(&self.cnx, &header)
+            .await?;
         let imp = Arc::clone(&self.imp);
 
         Request::spawn(
             "Access::AccessDialog",
             &self.cnx,
+            Arc::clone(&self.caller_authorization),
             handle.clone(),
             Arc::clone(&self.imp),
             Arc::clone(&self.spawn),
--- a/client/src/backend/caller.rs
+++ b/client/src/backend/caller.rs
@@ -1,0 +1,136 @@
+use zbus::{
+    message::Header,
+    names::{BusName, OwnedWellKnownName, WellKnownName},
+};
+
+use crate::{PortalError, backend::Result};
+
+/// Controls which D-Bus peers may call a portal backend interface.
+#[derive(Clone, Debug)]
+pub struct CallerAuthorization {
+    allowed_names: Option<Vec<OwnedWellKnownName>>,
+}
+
+impl CallerAuthorization {
+    /// Allow calls from any D-Bus peer.
+    pub fn allow_all() -> Self {
+        Self {
+            allowed_names: None,
+        }
+    }
+
+    /// Only allow the current owner of `well_known_name`.
+    pub fn require_name<'a, W>(well_known_name: W) -> zbus::Result<Self>
+    where
+        W: TryInto<WellKnownName<'a>>,
+        <W as TryInto<WellKnownName<'a>>>::Error: Into<zbus::Error>,
+    {
+        Self::require_names([well_known_name])
+    }
+
+    /// Only allow the current owner of one of `well_known_names`.
+    pub fn require_names<'a, I, W>(well_known_names: I) -> zbus::Result<Self>
+    where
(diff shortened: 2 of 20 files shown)


client/src/backend/secret.rs
```diff
@@ -74,3 +85,50 @@ impl SecretInterface {
         .await
     }
 }
+
+async fn ensure_frontend_caller(cnx: &zbus::Connection, header: &Header<'_>) -> Result<()> {
```
bilelmoussaoui: This should be done for all the portals though, no?

domenkozar: I audited the backend interfaces currently implemented by ashpd against the current `xdg-desktop-portal` call sites.

For the standard `xdg-desktop-portal` frontend, the current owner of `org.freedesktop.portal.Desktop` is the expected caller for:

* `Access`, `Account`, `AppChooser`, `Background`, `Email`, `FileChooser`, `Print`, `ScreenCast`, `Screenshot`, `Secret`, `Settings`, `Usb`, and `Wallpaper`
* backend `Request::Close` and `Session::Close`

Those calls originate in the `desktop-portal` process. The main methods also trust frontend-supplied application IDs, handles, and options; the request/session methods let the caller terminate in-flight work. ([Secret dispatch example](https://github.com/flatpak/xdg-desktop-portal/blob/8b8138a9651a54903daee1ade31d9b5634cd8042/desktop-portal/secret.c#L115-L161), [Request close](https://github.com/flatpak/xdg-desktop-portal/blob/8b8138a9651a54903daee1ade31d9b5634cd8042/desktop-portal/xdp-request.c#L53-L78), [Session close](https://github.com/flatpak/xdg-desktop-portal/blob/8b8138a9651a54903daee1ade31d9b5634cd8042/desktop-portal/xdp-session.c#L176-L185))

There are two interfaces that should not receive a hard-coded Desktop-owner check:

* `PermissionStore` has legitimate callers outside the Desktop frontend, including [`xdg-document-portal`](https://github.com/flatpak/xdg-desktop-portal/blob/8b8138a9651a54903daee1ade31d9b5634cd8042/document-portal/document-portal.c#L1899-L1902) and the [PipeWire permission manager](https://github.com/flatpak/xdg-desktop-portal/blob/8b8138a9651a54903daee1ade31d9b5634cd8042/pipewire/wp-plugin/xdp-wp-permission-manager.c#L259-L265). It needs a separate multi-caller policy.
* `Lockdown` needs authorization, particularly for its writable properties, but its specification says it is meant for [different portal frontends](https://github.com/flatpak/xdg-desktop-portal/blob/8b8138a9651a54903daee1ade31d9b5634cd8042/data/org.freedesktop.impl.portal.Lockdown.xml#L9-L26). Restricting it specifically to `org.freedesktop.portal.Desktop` would be too narrow.

So yes, this is broader than `Secret`. The clean ashpd shape looks like a reusable caller-authorizer applied to the frontend-only interfaces above, with explicit/configurable policies for `PermissionStore` and `Lockdown`, rather than an unconditional check across every `org.freedesktop.impl.portal.*` interface.

domenkozar: Reworked in 90761a1:

* added a shared `CallerAuthorization` policy, defaulting to the current owner of `org.freedesktop.portal.Desktop`
* applied it to the frontend-facing backend methods and propagated it to `Request::Close` / `Session::Close`
* gave `Lockdown` a separate configurable policy (inheriting the frontend policy by default) and authorized its property reads/writes
* gave `PermissionStore` a separate configurable policy that allows all callers by default, preserving its document-portal and PipeWire callers

The workspace tests/doctests, rustfmt, clippy with `-D warnings`, workspace check, and rustdoc with warnings denied all pass in the project CI container on Rust 1.92.0.

client/src/backend/access.rs
```diff
@@ -2,11 +2,13 @@ use std::sync::Arc;
 
 use async_trait::async_trait;
 use serde::{Deserialize, Serialize};
+use zbus::message::Header;
```
bilelmoussaoui, line 5: the include here is unnecessary and in all other files, it is only used once, no need for inclusion.

client/src/backend/caller.rs
```diff
@@ -0,0 +1,136 @@
...
+        let Some(header) = header else {
+            // Generated property-changed helpers invoke getters internally.
+            return Ok(());
+        };
+
+        self.authorize_fdo(connection, header).await
+    }
+}
+
+fn not_allowed() -> PortalError {
```
bilelmoussaoui, line 100: No need to have this as a helper here, if it is really repetitive, make it a method in ProtalError that is pub(crate)

client/src/backend/caller.rs
```diff
@@ -0,0 +1,136 @@
...
+    PortalError::NotAllowed("caller is not authorized to invoke this portal backend".into())
+}
+
+fn sender_matches_owner(sender: Option<&str>, owner: &str) -> bool {
```
bilelmoussaoui, line 104: Not sure why this has to be a method

bilelmoussaoui (commented):
overall, lgtm, but can't we add test for this?

client/src/backend/access.rs
```diff
@@ -117,15 +119,22 @@ pub(crate) struct AccessInterface {
     imp: Arc<dyn AccessImpl>,
     spawn: Arc<dyn futures_util::task::Spawn + Send + Sync>,
     cnx: zbus::Connection,
+    caller_authorization: Arc<CallerAuthorization>,
 }
 
 impl AccessInterface {
     pub fn new(
         imp: Arc<dyn AccessImpl>,
         cnx: zbus::Connection,
```
bilelmoussaoui, line 128: To simplify the addition of this caller_authorization, i would add a new struct in separate PR/commit, that holds the cnx and spawn field, something like IfaceInner. then adding caller_authorization wouldn't require changing a ton of places..

PR #488, opened 2026-09-17
proxy: disable zbus property caching for Request/Session objects

### LINKED ISSUE #487: # `Request` / `Session` proxies are built with property caching on — one `zbus::proxy` WARN per proxy on portals that don't serve `org.freedesktop.DBus.Properties` on those objects
**ashpd 0.13.13, zbus 5.19.0, xdg-desktop-portal 1.22.1, xdg-desktop-portal-hyprland 1.4.1** (Hyprland 0.56.2, Arch/CachyOS). The same shape on `xdg-desktop-portal-wlr` from reports I have seen; not reproduced there myself.

## What happens

Every `ScreenCast` / `RemoteDesktop` flow logs four WARN lines that come from zbus's property-caching task, not from the application:

```
WARN {}{task_name="org.freedesktop.portal.Session proxy caching"}: zbus::proxy: Failed to populate properties cache via GetAll: org.freedesktop.DBus.Error.UnknownMethod: L’objet n’existe pas à l’emplacement « /org/freedesktop/portal/desktop/session/1_5307/punktfunk… »
WARN {}{task_name="org.freedesktop.portal.Request proxy caching"}: zbus::proxy: Failed to populate properties cache via GetAll: … /org/freedesktop/portal/desktop/request/1_5307/…
WARN {}{task_name="org.freedesktop.portal.Request proxy caching"}: …
WARN {}{task_name="org.freedesktop.portal.Request proxy caching"}: …
```

(one `Session` + one `Request` per `create_session`, one `Request` per `select_sources`, one per `start`). On a streaming host that opens a portal session per client that is **116 lines in 4 hours** over 29 sessions, and they are the only WARNs in a normal run — `journalctl -p warning` becomes useless and every user report from a wlroots-family compositor carries them as a red herring. The calls themselves all succeed; nothing reads a property on either object.

## Cause

`desktop/request.rs:223` and `desktop/session.rs:44` build the transient-object proxies through `Proxy::new_desktop_with_path`, which lands in `proxy.rs:80`:

```rust
let inner: zbus::Proxy = zbus::proxy::Builder::new(&connection)
    .interface(interface)?
    .path(path)?
    .destination(destination)?
    .build()
    .await?;
```

No `.cache_properties(...)`, so zbus defaults to `CacheProperties::Yes` and spawns its `"{interface} proxy caching"` task (`zbus-5.19.0/src/proxy/mod.rs:280`), which calls `org.freedesktop.DBus.Properties.GetAll` on the object and logs at WARN when that fails (`proxy/mod.rs:301`). `xdg-desktop-portal-hyprland` does not implement the `Properties` interface on `/org/freedesktop/portal/desktop/request/*` and `/session/*` — the objects exist, `Response` signals arrive and `Close` works, but `GetAll` is `UnknownMethod`. The `version` read that follows in `with_connection` already tolerates this (`Err(_) => Ok(1)`), so ashpd itself is fine; only the background task complains.

Neither `Request` nor `Session` exposes a property that a caller could read through ashpd's API — `Request` is only `Response` + `Close`, `Session` is `Close` + `Closed` — so the cache can never be used on these two proxies. The task is pure overhead: one extra round trip per proxy, plus a WARN on every portal that does not serve `Properties` there.

## Proposed fix

Build those two proxies with caching off. Smallest form, keeping `new_desktop_with_path` for everyone else:

```rust
// src/proxy.rs
pub(crate) async fn new_desktop_with_path_uncached<P>(
    connection: zbus::Connection,
    interface: &'a str,
    path: P,
) -> Result<Proxy<'a>, Error>
where
    P: TryInto<ObjectPath<'a>>,
    P::Error: Into<zbus::Error>,
{
    let inner: zbus::Proxy = zbus::proxy::Builder::new(&connection)
        .interface(interface)?
        .path(path)?
        .destination(DESKTOP_DESTINATION)?
        .cache_properties(zbus::proxy::CacheProperties::No)
        .build()
        .await?;
    // No `version` read: these objects have no properties at all.
    Ok(Self { inner, version: 1 })
}
```

and use it from `desktop/request.rs:223` and `desktop/session.rs:44`. Skipping the `version` GET on these two also removes a second `GetAll`/`Get` round trip that can only ever fail on them.

Alternatively a `cache_properties` parameter on `with_connection`, defaulting to `Yes`, if a wider API is preferred.

## Repro

Any `ScreenCast` flow with `RUST_LOG=info` on a Hyprland desktop:

```rust
let proxy = ashpd::desktop::screencast::Screencast::new().await?;
let session = proxy.create_session().await?;          // → Session WARN + Request WARN
proxy.select_sources(&session, CursorMode::Embedded, SourceType::Monitor.into(), false, None, PersistMode::DoNot).await?;   // → Request WARN
proxy.start(&session, None).await?;                   // → Request WARN
```

Downstream: punktfunk silenced it on their side with `info,zbus::proxy=error` in the default `EnvFilter` (https://git.unom.io/unom/punktfunk/issues/1136), which also hides any *other* `zbus::proxy` WARN — the proper fix is here.

Klos54:
Fixes #487.

`Request` and `Session` proxies were built through `Proxy::new_desktop_with_path` → `with_connection` with zbus's default `CacheProperties::Yes`. zbus then spawns its `"{interface} proxy caching"` task, which calls `org.freedesktop.DBus.Properties.GetAll` on the object and logs at WARN when that fails. Portals that do not serve the `Properties` interface on `/org/freedesktop/portal/desktop/request/*` and `/session/*` (xdg-desktop-portal-hyprland, -wlr) answer `UnknownMethod`, so every `Request` and `Session` produces one WARN while the calls themselves succeed.

Neither object exposes a property that is read through this crate (`Request` is `Response` + `Close`, `Session` is `Close` + `Closed`; neither wrapper has a `version()` accessor), so the cache can never be used on them. `new_desktop_with_path` is only called from `desktop/request.rs` and `desktop/session.rs` and `mod proxy` is private, so this changes it in place: build with `CacheProperties::No` and skip the `version` property read, which fails the same way on those portals.

Checked on Hyprland 0.56.2 / xdg-desktop-portal 1.22.1 / xdg-desktop-portal-hyprland 1.4.1 with a `Screencast::create_session` + `close` run under `RUST_LOG=info`: two `zbus::proxy` WARNs (`Request proxy caching`, `Session proxy caching`) before, none after; `cargo fmt --check` clean, clippy unchanged.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


client/src/proxy.rs
```diff
@@ -106,6 +106,14 @@ impl<'a> Proxy<'a> {
         Ok(Self { inner, version })
     }
 
+    /// Build a proxy for a transient portal object (`Request` / `Session`).
```
bilelmoussaoui: The comment is too long for no reason, please trim it down to the bare minimum

Klos54: Trimmed to two lines.

client/src/proxy.rs
```diff
@@ -115,7 +117,14 @@ impl<'a> Proxy<'a> {
...
     {
-        Self::with_connection(connection, interface, path, DESKTOP_DESTINATION).await
+        let inner: zbus::Proxy = zbus::proxy::Builder::new(&connection)
+            .interface(interface)?
+            .path(path)?
+            .destination(DESKTOP_DESTINATION)?
+            .cache_properties(zbus::proxy::CacheProperties::No)
+            .build()
+            .await?;
+        Ok(Self { inner, version: 1 })
```
bilelmoussaoui: I would still read the version because it might eventually change in the future.

Klos54: Done — the `version` read is back. Building moved into a private `build()` taking the `CacheProperties`, so `with_connection` and `new_desktop_with_path` share the same version-probing code and only the cache flag differs.

